PR #27, opened 2022-08-01
`Metadata` adjustments and more Algorithms

matheus23:
* Pull out `type` and `version` fields from `Metadata
* Make `Metadata` more obviously extensible

### DIFF
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -69,7 +69,10 @@ type PrivateNodeHeader = {
 }
 
 type PrivateDirectory = {
-  metadata: Metadata<"wnfs-private-dir">
+  type: "wnfs/priv/dir"
+  version: "0.2.0"
+  // userland:
+  metadata: Metadata
   entries: Record<string, {
     contentKey: Key // hash(deriveKey(entryRatchet))
     revisionKey: Encrypted<Key> // encrypt(deriveKey(ratchet), deriveKey)(entryRatchet))
@@ -80,7 +83,10 @@ type PrivateDirectory = {
 }
 
 type PrivateFile = {
-  metadata: Metadata<true>
+  type: "wnfs/priv/file"
+  version: "0.2.0"
+  // userland:
+  metadata: Metadata
   content: ByteArray | {
     ratchet: SkipRatchet
     count: Uint64
--- a/spec/public-wnfs.md
+++ b/spec/public-wnfs.md
@@ -16,9 +16,11 @@ type PublicNode
   | PublicFile
 
 type PublicDirectory = {
-  previous?: CID<CBOR<PublicDirectory>>
-  merge?: CID<CBOR<PublicDirectory>>
-  metadata: Metadata<"wnfs-dir">
+  type: "wnfs/pub/dir"
+  version: "0.2.0"
+  previous: Array<CID<CBOR<PublicDirectory>>>
+  // userland:
+  metadata: Metadata
   entries: Record<string, CID<CBOR<PublicNode>> | PublicSymlink>
 }
 
@@ -27,29 +29,21 @@ type PublicSymlink = {
 }
 
 type PublicFile = {
-  previous?: CID<CBOR<PublicFile>>
-  merge?: CID<CBOR<PublicFile>>
-  metadata: Metadata<"wnfs-file">
+  type: "wnfs/pub/file"
+  version: "0.2.0"
+  previous: Array<CID<CBOR<PublicFile>>>
+  // userland:
+  metadata: Metadata
   content: CID<IPFSUnixFSFile>
 }
 ```
 
 ## Metadata
 
-Metadata in public WNFS must be extensible. Additional fields in the metadata type must be parsed without errors and preserved between versions unless explicitly modified by an application.
+Metadata in WNFS is a CBOR record with arbitrary keys and values.
 
-```typescript
-type Metadata<Type> = {
-  version: "0.2.0"
-  type: Type
-  unixMeta: UnixMeta
-}
-
-type UnixMeta = {
-  mtime: Int // POSIX timestamp
-  ctime: Int // POSIX timestamp
-  mode: Int // POSIX file mode, e.g. (octal) 666 or 777
-}
+```ts
+type Metadata = CBOR<Record<string, any>>
 ```
 
 ## Algorithms


.github/workflows/words-to-ignore.txt
```diff
@@ -50,6 +50,8 @@ AES-256
 CBOR like ABC
 distinguishability
 bitarray like noun
+cryptree like noun
```
expede, line 53: Oh cooooool

matheus23: I should probably change a bunch of them to be `bitarray like array` and `cryptree like tree`.
This helps "copy" the rules for prefixes and suffixes from the word that they're alike to. `noun` here may be really confusing, but it's literally copying the prefixes and suffixes for the word "noun".

matheus23:
I essentially 'reverted' to putting the `PrivateNodeHeader` encrypted within `PrivateDirectory` and `PrivateFile` again.
Previously, our HAMT (at the encrypted layer) had this type: `HAMT<Namefilter, [Ciphertext, Array<CID<Ciphertext>>]`, but now it's `HAMT<Namefilter, Array<CID<Ciphertext>>`.
The reason is that not all blocks in the private file system have a `PrivateNodeHeader`: When files are sharded, then the shards don't have headers. And we don't want to leak the metadata about which values in the HAMT are for file shards and which are for nodes themselves.

So now by putting the node headers within the nodes, we get back a single ciphertext per node-and-header.

spec/skip-ratchet.md
```diff
@@ -30,3 +30,38 @@ function deriveKey(ratchet: SkipRatchet): ByteArray<32> {
...
 ```
 
+### Increasing
+
+The skip ratchet can be stepped forward, but not backwards.
+
+Forward stepping can happen at
+- the small digit, in increments of one,
+- the medium digit, in increments of (mostly) 256,
+- the large digit, in increments of (mostly) 65536.
```
expede, line 40: The word "mostly" here reads as a hedge. If you mean "of at most" (because of the zeroing on the smaller digits), perhaps rephrase to something like...

```suggestion
- the small digit, in increments of one
- the medium digit, in increments up to 256
- the large digit, in increments up to 65536

These intervals for each digit are often referred to as small, medium or large "epochs". The reason that it isn't exactly 256 and 65536 is because each ratchet can only be incremented to the beginning of their next epoch.
```

expede: I would also suggest either including this diagram as a screenshot...

Screen Shot 2022-08-10 at 15 02 22

...or doing an ASCII version. This seems to be the image that gives folks the biggest "a-ha"s!

spec/public-wnfs.md
```diff
@@ -16,9 +16,11 @@ type PublicNode
   | PublicFile
 
 type PublicDirectory = {
-  previous?: CID<CBOR<PublicDirectory>>
-  merge?: CID<CBOR<PublicDirectory>>
-  metadata: Metadata<"wnfs-dir">
+  type: "wnfs/pub/dir"
+  version: "0.2.0"
```
expede, line 20: Love it 👍 

Arguably we can duck type the directory, but 18 chars per header is fine by me

spec/public-wnfs.md
```diff
@@ -27,32 +29,46 @@ type PublicSymlink = {
...
+  // userland:
+  metadata: Metadata
   content: CID<IPFSUnixFSFile>
 }
 ```
 
 ## Metadata
 
-Metadata in public WNFS must be extensible. Additional fields in the metadata type must be parsed without errors and preserved between versions unless explicitly modified by an application.
+Metadata in WNFS is a CBOR record with arbitrary keys and values.
```
expede, line 43: "Record" implies having specific fields (to me at least). Perhaps "map" instead (that's the term they use in the CBOR spec)

```suggestion
The metadata field MUST be a CBOR map. It is in userland, and contains arbitrary keys and values.
```

spec/public-wnfs.md
```diff
@@ -27,32 +29,46 @@ type PublicSymlink = {
 }
 
 type PublicFile = {
-  previous?: CID<CBOR<PublicFile>>
-  merge?: CID<CBOR<PublicFile>>
-  metadata: Metadata<"wnfs-file">
+  type: "wnfs/pub/file"
+  version: "0.2.0"
```
expede, line 33: 🥳

spec/private-wnfs.md
```diff
@@ -12,22 +12,23 @@ It makes sense to split the private partition into two layers:
```
expede, line 12: Recommendation: adding a legend for `CK`, `SK`, and the dashed line would be helpful here.

spec/private-wnfs.md
```diff
@@ -12,22 +12,23 @@ It makes sense to split the private partition into two layers:
 
 ## The Encrypted Layer
 
-The private partition consists of lots of private data blocks encrypted with different keys. These blocks are supposed to be of size smaller than 256 kilobytes in order to comply with block size restrictions from IPFS. However, keeping block size small is also useful for reducing metadata leakage - it's less obvious what the file size distribution in the private file system is like, if these files are split into blocks.
+Unlike in the public partition, there is no "root" in the private partition. The private file system can contain multiple unrelated private trees at the same time. Thus, we refer to it as a `PrivateForest`.
```
expede, line 15: We probably want to be very clear here. "Layer" and "root" can have several meanings. There _is_ a root at the data layer, and at least one root at the file layer.

We can go pretty deep in clarifying these layers, but in interest of time for now I'd perhaps clarify something like:

```suggestion
The private file system's encrypted "root" is not the root of the decrypted file system.

The private file system can contain multiple unrelated private trees at the same time. Thus, we refer to it as a `PrivateForest`.
```

BTW, I would like to get a Dark Forest paper on arXiv ahead of StrangeLoop. We'll see if that's even a realistic timeline, but I would love your input or to coauthor such a paper.

spec/private-wnfs.md
```diff
@@ -12,22 +12,23 @@ It makes sense to split the private partition into two layers:
...
+Unlike in the public partition, there is no "root" in the private partition. The private file system can contain multiple unrelated private trees at the same time. Thus, we refer to it as a `PrivateForest`. 
 
-These blocks of encrypted data are put into a HAMT that encodes a multi-valued hash-map. The HAMT has a node-degree of 16. See [`rationale/hamt.md`](/rationale/hamt.md) for more information about that.
+At the encrypted layer, the private forest is a bunch of private data blocks encrypted with different keys. These blocks are supposed to be of size smaller than 256 kilobytes in order to comply with default block size restrictions from IPFS. However, keeping block size small is also useful for reducing metadata leakage - it's less obvious what the file size distribution in the private file system is like, if these files are split into blocks.
```
expede, line 17: No changes required for now; this text is fine as is for our implementation. However, I would suggest reading [RFC2119](https://www.rfc-editor.org/rfc/rfc2119) (it's short!), and using that language. It does change the way you construct sentences, but does help ideas read very clearly and with a clear level of how important a certain detail is.

For example:

> These blocks are SHOULD be less than 256 kilobytes in order to comply with default block size restrictions from IPFS.

or 

> Keeping blocks under 256kb is RECOMMENDED in order to comply with the default block size in IPFS

spec/private-wnfs.md
expede: > - The "decrypted" layer that defines the type of data you can decrypt given the correct keys. Links between blocks in this layer are references in the HAMT data structure at the "encrypted" layer.
> - The "encrypted" layer that defines how all of the encrypted data blocks are organized as IPLD data. Links in this layer are CID-links.

These are often called the "data" and "file" semantic levels. I agree that it's important to clarify that these are encrypted/decrypted in this context, but referencing these layers (or reversed: using these terms and clarifying that files are encrypted) may be a useful since they're existing terms of art. Many file systems will also talk about a "physical" layer, but we're not laying data out on disk.

spec/private-wnfs.md
```diff
@@ -96,7 +108,71 @@ Example:
...
+
+To resolve a namefilter hash, look up the hash in the HAMT. The resulting key-value pair will give you the full namefilter as well as the private node header and a list of at least one private directory.
+
+Looking up a namefilter hash in the HAMT works by splitting a hash into its nibbles. For example: a hash `0xf199a877d0...` gets split into the nibbles `0xf`, `0x1`, `0x9`, etc.
+
+The nibble order is first taking the 4 most significant bits of and *then* the 4 least significant bits for each hash byte from byte index 0 to 31. This way this matches the usual hex encoding of byte-strings and reading the hex digits off one-by-one.
+
+Each nibble is used as an identifier for a `Node`s child `Node`. Starting at the root, find the root `Node`s child by taking the nibble and computing the index of the child node as
+
+$$index = popcount(bitmask \land ((1 \ll nibble) - 1))$$
```
expede, line 127: Fine as-is, but might I suggest using a different font for function names:

```suggestion
$$\textsf{index} = \textsf{popcount}(bitmask \land ((1 \ll nibble) - 1))$$
```
Screen Shot 2022-08-10 at 15 29 45

spec/private-wnfs.md
```diff
@@ -96,7 +108,71 @@ Example:
...
+
+If the child is a Node, repeat the process of with the next nibble.
+
+If the child is a HAMT bucket of values, iterate that bucket to find one that has a namefilter that matches the hash of the namefilter. The associated values then contains the ciphertexts and the algorithm is done.
+
+
+### Private Versioning
+
+`: (Namefilter, RevisionKey) -> Namefilter`
```
expede, line 136: Interesting; is the `:` with nothing to the left a Rust convention?

spec/private-wnfs.md
```diff
@@ -96,7 +108,71 @@ Example:
...
+- $inc$ refers to the [skip ratchet increase operation](/spec/skip-ratchet.md#Increasing)
+
+It is possible to choose $n$ in $inc^n(ratchet)$ and due to the properties of the skip ratchet skip ahead versions instead of having to skip one-by-one. When looking for the most recent version of a file, we recommend first skipping to the next small epoch, then the next medium, then large epochs, as long as these revisions exit, and then backtracking once an unpopulated revision is found.
+
+
+### Path Resolving
+
+`: (PrivateDirectory, Array<string>) -> Hash<Namefilter>`
+
+Resolving paths in the private file system is always relative. The private partition is actually a forest of trees, and you may not know what other cryptrees exist in there next to what an app may consider the "root". We sometimes refer to the private forest as a "dark forest".
```
expede, line 161: Clarify "relative to what?"

expede: > We sometimes refer to the private forest as a "dark forest".

Let's just make this official somewhere near the top of the document  — it's too good to let pass 😁

expede (approved):
Just some notes, but looking pretty great overall! Nothing substantive to change, just nits, hence the `Approved` 🚀

spec/skip-ratchet.md
matheus23: Very good idea! Added.

spec/public-wnfs.md
matheus23: Yeah, I think that's clearer. I'm used to bad naming from typescript (`Record<string, string>` is a map... because javascript).

spec/private-wnfs.md
matheus23: Hmm. Okay, back when I read the whitepaper, I found the "data" vs. "file" layer distinction confusing.

I think there's already a "data" layer, in, let's say, the public filesystem already, that we're not talking about. That is the underlying IPLD that a file system gets serialized to, which doesn't quite map to file systems themselves. The file layer abstracts over the UnixFS directory shard and the UnixFS file chunking.

I guess in the whitepaper in a way there already existed multiple "data" layers :sweat_smile: 

I'd be in favor of reserving the "data" keyword for another layer I may want to describe in the future. Does that sound reasonable?

spec/private-wnfs.md
matheus23: Oh yeah, I see how that can be confusing!
Great suggestions, I'm grabbing that :3

spec/private-wnfs.md
matheus23: I want to do a spec-pass over the whole spec with that in mind in the future. I just read the RFC, it's really clear, and MUST/REQUIRED vs. SHOULD/RECOMMENDED vs. MAY makes a lot of sense to me. Thanks!

spec/private-wnfs.md
matheus23: :man_cook: :ok_hand:

spec/private-wnfs.md
matheus23: No, It's my own convention :sweat_smile: 

Initially it was
```md
### Private Versioning `: (Namefilter, RevisionKey) -> Namefilter`
```

But that looked horrible when rendered, and it produced really unpredictable anchor ids. So I just pushed it to the next line.

I should clarify that in `notation.md`. Do you like/dislike it? Does it help?

spec/private-wnfs.md
expede: > I'd be in favor of reserving the "data" keyword for another layer I may want to describe in the future, specifically all the sharding & chunking. Does that sound reasonable?

Ah, indeed: you want to be able to describe the decrypted data layout as a separate later from the stored encrypted bytes and IPLD layout. (e.g. We probably don't need as much sharding and chunking because it's just big CBOR maps rather than IPLD.) Okay yeah — having separate names for these is a good idea 💯 You're going to laugh that I'm about to suggest YET ANOTHER stack diagram 😛 Something like this (or however you're thinking about it) may be helpful:

```
┌────────────────┬──────────────────┐
│                │                  │
│                │       File       │
│                │                  │
│   Decrypted    ├──────────────────┤
│                │                  │
│                │       Data       │
│                │                  │
├────────────────┼──────────────────┤
│                │                  │
│   Encrypted    │       Data       │
│                │                  │
└────────────────┴──────────────────┘
```

Which implies an "encrypted data" and "decrypted data" layers, plus an implied "decryption" to the file layer...

Wait LOL this is why this kind of exercise is useful. I'm going leave the above text in place even. It needs more cleanup than below, but there's almost more like separate dimensions:

| | Ciphertext | Cleartext |
|--|-------|-------------|
| Data | Block Layout | CBOR Schema |
| File | Namefilters & multivalues | Files & dirs |

matheus23: I like the stack diagram :) Keeping that idea for once I introduce the "data layer" concept.

spec/private-wnfs.md
matheus23: I added the 'dark forest' reference to where I'm introducing the 'forest' property itself, below the decrypted layer heading.

PR #32, opened 2022-08-23
Clarifications on Private FS Structure

expede:
[Preview 📜](https://github.com/wnfs-wg/spec/blob/expede/clarifications/spec/private-wnfs.md)


matheus23:
image

In the temporal header for "Documents/" it shows "Skip Ratchet", but I'd like to clarify that should probably be "Node Key" (for the Documents node).

expede:
> In the temporal header for "Documents/" it shows "Skip Ratchet", but I'd like to clarify that should probably be "Node Key" (for the Documents node).

Ah good! This was actually on my backlog of questions :)

spec/private-wnfs.md
```diff
@@ -1,80 +1,117 @@
...
 }
 
-type Node<K, V> =
-  [ ByteArray<2> // bitmask
-  , Array<Entry<K, V>> // Entries
-  ]
+type Node<K, V> = [
+  ByteArray<2>, // Bitmask Key
+  Array<Entry<K, V>> // Entries
+]
```
expede, line 72: Just normalizing this to TypeScript... unless that community has adopted the ~~One True Way~~ pretty flush-left comma layout, of course 🤞

spec/private-wnfs.md
```diff
@@ -1,80 +1,117 @@
...
+┌─────────┼────────────────────────Bucket───────────────────────────────────────┐
+│         ╷                                                                     │
+│  ┌──────┴───────┐  ┌──────────────┐  ┌──────────────┐  ┌──────────────┐       │
+│  │              │  │              │  │              │  │              │       │
+│  │   Expanded   │  │  Ciphertext  │  │  Ciphertext  │  │  Ciphertext  │  ...  │
+│  │  Namefilter  │  │    Node A    │  │    Node B    │  │    Node C    │       │
+│  │              │  │              │  │              │  │              │       │
+│  └──────────────┘  └──────────────┘  └──────────────┘  └──────────────┘       │
+│                                                                               │
+└───────────────────────────────────────────────────────────────────────────────┘
```
expede, line 103: I feel really bad cutting the pretty colourful images 😭 I'm not fully opposed to SVG exporting them out of Figma, but also I'm `n` instances of having to remove my own beautiful colourful images from documents that I'm pretty much just at "whatever the platform supports natively"

spec/private-wnfs.md
```diff
@@ -1,80 +1,117 @@
...
 
-The private file system's encrypted "root" is not the root of the decrypted file system.
+These all form graphs, where the nodes and links have different meanings per layer. The 
 
-A single file system's encrypted root can represent a whole forest of decrypted file system trees. The roots of these trees may be completely unrelated. We refer to them as the `PrivateForest`. Since you may not know what else there is in the forest, we sometimes refer to the it as the "dark forest".
+| Visibility | Layer | Node        | Link             |
+|------------|-------|-------------|------------------|
+| Decrypted  | File  | WNFS File   | File Path        |
+| Decrypted  | Data  | CBOR Object | Namefilter + Key |
+| Encrypted  | Data  | IPLD Block  | CID              |
```
expede, line 40: I still feel like there's more that we could do to pull these layers apart 🤔 Nothing to do immedietly, more of an intuition that these layers really do blur into each other a bit at the moment.

spec/private-wnfs.md
```diff
@@ -112,57 +149,264 @@ type PrivateFile = {
...
   header: Encrypted<CBOR<PrivateNodeHeader>>
-  // userland:
+
+  // USERLAND
   metadata: Metadata
-  content: ByteArray | {
-    ratchet: SkipRatchet
+  content: ByteArray | { // FIXME what is the SR here for? Is this a tombstone? If so, is that not a different node type?
+    ratchet: SkipRatchet 
     count: Uint64
```
expede, line 157: @matheus23 what's the `| {ratchet, count}` in here? My two guesses:

* Reusing the layout for header nodes
* It's a pointer to another version (e.g. tombstone + signpost)

expede:
The remaining spellcheck complaints are the FIXMEs in the text. I still need to give this a top-to-bottom read on the expanded version, but I _think_ this is ready to go.

spec/private-wnfs.md
matheus23: I feel bad about that too :(
Not to insult these carefully crafted ASCII images, but they read worse than images/SVGs. There's just way fewer tools to convey information in ASCII art, e.g. there's no different colors for arrows.
So it ends up being a trade-off: It's worse for readers, but better for contributors.
I'd argue we should optimize for readers rather than contributors at this point, especially since I doubt there's going to be contributions to the key structure diagrams themselves.
Maybe we can help contributors out by adding a `CONTRIBUTING.md` that explains the process needed for adjusting these images: Message me/someone from fission and either request a change or request access to the figjam board. Or just build another image with whatever tools the contributor feels comfortable with.

spec/private-wnfs.md
matheus23: It's for finding the private file shards: https://whitepaper.fission.codes/file-system/partitions/private-directories/platform-layer#content-access

spec/private-wnfs.md
expede: > Not to insult these carefully crafted ASCII images

No worries at all — I wouldn't call them "carefully crafted". I just slapped them together in [ASCIIFlow](https://asciiflow.com/). They did not take a ton of time aside from fighting GitHub's rendering engine.

I agree that they're "worse" than the colourful version, but they read _sufficiently well to understand what's happening._ I've also needed to make a bunch of changes while clarifying the structure in the diagrams, so having them be editable has already been pretty important. There's a reason why ASCII art is so ubiquitous — it's extremely portable.

Even in my adding clarifications, I needed to adjust some of the diagram layouts (whether in ASCII or Figma).

> Message me/someone from fission and either request a change or request access to the figjam board

That is a very high bar / roadblock for contribution. We should avoid this as a pattern, since this is a community project. Why does Fission own all of the tools for this? It's now also gated on the handful of people that have access to this Figjam, the diagrams disappear if Fission stops using Figma, etc.

@matheus23 can you [export the Figjam to SVG](https://help.figma.com/hc/en-us/articles/360040028114-Guide-to-exports-in-Figma) and let's explore how painful it is to edit in e.g. Inkscape?

QuinnWilton: I agree with Brooke that lowering the bar for contributing is an important goal, especially for specs, and this is something I've been grappling with myself in trying to find the right balance between ascii and mathjax for some of the dialog specs.

One possibility that I don't think any of us have discussed is using DOT to define these graphs, with a GitHub action to build the PNGs and commit them to the repo. That could happen right on a PR's branch if the relevant file changes, so that the preview even updates automatically. With a nice looking Fission theme for Graphviz, that might give us the best balance between expressivity and barrier to entry.

expede: That's an interesting approach 🤔 We should definitely play with that

spec/private-wnfs.md
```diff
@@ -112,57 +149,274 @@ type PrivateFile = {
...
-  content: ByteArray | {
-    ratchet: SkipRatchet
-    count: Uint64
-  }
+  content: ByteArray | ExternalContent
+}
+
+type ExternalContent = {
+  ratchet: SkipRatchet 
+  count: Uint64
```
expede, line 160: CC @matheus23

spec/private-wnfs.md
```diff
@@ -172,34 +426,28 @@ Where
...
+`resolvePath : (PrivateDirectory, Array<string>) -> Hash<Namefilter>`
 
-`: (PrivateDirectory, Array<string>) -> Hash<Namefilter>`
+Paths in the private file system MUST be resolved relative to a parent directory. This directory MAY be at the current revision, or MAY be at an earlier version.
 
-Paths in the private file system need to be resolved relative to a directory to resolve the path from.
+Path resolution can happen in three modes:
+1. Resolve the current snapshot: only resolve the current snapshot of a version. This only requires a content key for decryption.
+2. Seek: For each path segment, look up the most recent version that can be found (as described in the [private versioning algorithm](#Private-Versioning)). This requires access to the directory's revision key.
+3. Seek & Repair: Similar to seeking, this mode searches for the latest version of a file. Once it is found, and in the case that it's not the same as what the parent refers to, it'll repair the parent's link to link to the most recent revision. FIXME BFT dangers? Defensive programming?
```
expede, line 440: Ah right; I think that we may not be able to get around this, and I think we discussed this before... but also just wanted to call it out here so that we're aware of the tradeoff of defensive programming

CC @QuinnWilton any thoughts?

spec/private-wnfs.md
```diff
@@ -1,80 +1,117 @@
...
 
 type Entry<K, V>
-  = CID<CBOR<Node<K, V>>>
-  | Array<[K, V]> // bucket of values
+  = CID<CBOR<Node<K, V>>> // Subentry sharding
+  | Array<[K, V]> // Bucket of values
 ```
 
-## The Decrypted Layer
+In the above, note that `Node<K, V>` and `Entry<K, V>` are mutually recursive types. This permits the sharding of entries beyond the default IPLD link limit. FIXME @matheus plz confirm.
```
expede, line 79: @matheus23 Just double checking that this new text this correct?

spec/private-wnfs.md
matheus23: > Even in my adding clarifications, I needed to adjust some of the diagram layouts (whether in ASCII or Figma).

Yeah, my argument here was, if it's mostly you and me doing changes, sharing a FigJam board is possible for us.

> Why does Fission own all of the tools for this? It's now also gated on the handful of people that have access to this Figjam, the diagrams disappear if Fission stops using Figma, etc.

I don't think that's true: The diagrams are exported and pretty easy to recreate. If Fission stops using Figma, the images are still there.
It's also possible to export FigJam files ([here's the one I used for creating these diagrams](https://ipfs.runfission.com/ipns/matheus23.files.fission.name/p/Pics/Private%20WNFS%20v2%20Key%20Hierarchy.jam)). I know it's not like FigJam files are an open standard. I'm only suggesting this as an option to making figjam file changes easier for outside contributors without having to invite them to some Fission Figma workspace.

> can you [export the Figjam to SVG](https://help.figma.com/hc/en-us/articles/360040028114-Guide-to-exports-in-Figma) and let's explore how painful it is to edit in e.g. Inkscape?

Unfortunately only Figma supports SVG exports. FigJam doesn't. I think .jam file exports may be better (even though they're not standardized).

---

The images being made using FigJam doesn't keep anybody from being able to contribute diagrams. They just need to create a PR with the image committed to the repository.
I view this similarily as people contributing logos to open source projects: They contribute images/svgs. That's fine even if they used proprietary tools to create these graphics. I know that these diagrams are not quite the same as logos, they *do* change more often, but maybe not often enough to warrant being write-optimized over read-optimized?

Another data point is the [Signal protocol](https://signal.org/docs/specifications/doubleratchet/#kdf-chains). This is where I went looking originally for inspiration/best practices when it comes to diagrams.

This is not a "over my dead body"! Just raising these points before we make a decision.

spec/private-wnfs.md
matheus23: I didn't know the term [Defensive Programming](https://en.wikipedia.org/wiki/Defensive_programming) before.
> Overly defensive programming, however, may safeguard against errors that will never be encountered, thus incurring run-time and maintenance costs.

I want to stress that this type of seek & repair is *required* for WNFS to function correctly. This is what you've previously called "collaborative rooting". Maybe I should include that term there :thinking:

spec/private-wnfs.md
matheus23: Yes? Kind of?
This part is describing the internal node structure of the HAMT.
These nodes will never come even close to the the blocksize limit. (Assuming that's what you were referring to. What is the IPLD link limit?)

The Entry buckets will actually never have more than 3 entries. That's the Filecoin HAMT's chosen bucket size and what's performing quite well.

spec/private-wnfs.md
expede: Oooooh I see. I think we should change the wording there. Will propose a change. Thanks for the clarification 👍

spec/private-wnfs.md
expede: I think this conversation is closed as Graphviz + GH Action seems to meet our needs. Please do reopen if you want to continue the conversation @matheus23 :)

spec/private-wnfs.md
expede: haha okay. So to repeat that back: "These are just the easiest types to write. There's a hard limit in our existing implementation due to a dependency, but we don't expect to hit it. There's no hard theoretical limit here, but it is the sensible choice in practice." Is that roughly correct?

matheus23: > "There's no hard theoretical limit here, but it is the sensible choice in practice."

I'm assuming with limit you're referring to "a limit on the number of bucket entries".

Yeah. There's none. But the bucket limit *must* be set to 3 in all WNFS implementations. Otherwise we're losing history-independence and some useful properties that come with it, like having CIDs be the same so we can cut-off computations.
Yeah, the bucket size limit was missing from the spec before this.

---

I'm still a little confused by the conversation though. What is the IPLD link limit?
In the text above I'm talking about the buckets, which are the leafs of the tree, not the recursive Node links.

The recursive node links are *always* capped at 16, because that's the degree of our HAMT, and that's unrelated to IPFS block size or anything like that.

spec/private-wnfs.md
```diff
@@ -1,80 +1,82 @@
...
-## The Decrypted Layer
+In the above, note that `Node<K, V>` and `Entry<K, V>` are mutually recursive types. This permits the sharding of entries beyond the default IPLD link limit.
+
+## 2.2 Ciphertext Files
 
-The private WNFS borrows the same metadata structure as the [public WNFS](/spec/public-wnfs.md#metadata).
+The encrypted file layer is a very thin enrichment of the data layer. In particular, it knows about about namefilters as labels, and ciphertext blobs as being separate from the expanded namefilter inside the bucket.
 
-SHA3 hashes of namefilters are the linking scheme used in the decrypted layer.
+![](./diagrams/hamt_leaves.svg)
```
matheus23, line 69: I think the label "Bucket" here is misleading: The HAMT buckets contain multiple key-value pairs (up to 3). Maybe a more appropriate label would be "Entry"?

spec/private-wnfs.md
```diff
@@ -112,57 +114,126 @@ type PrivateFile = {
   version: "0.2.0"
   // encrypted using deriveKey(ratchet)
   header: Encrypted<CBOR<PrivateNodeHeader>>
-  // userland:
+
+  // USERLAND
```
matheus23, line 118: For consistency, maybe we change the `// userland:` comment in the `PrivateDirectory` definition as well?

spec/private-wnfs.md
```diff
@@ -112,57 +114,126 @@ type PrivateFile = {
...
 
+### 3.2.1 Node Headers
 
-Example:
+Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to [§3.2.3 Pointers & Keys](#323-pointers--keys) for more detail.
 
-![block encryption example](/images/encrypted_blocks.png)
+### 3.2.2 Node Metadata
 
+Node metadata is the userland equivalent of the node's header.
```
matheus23, line 135: Should we merge this sentence with the one above:

> The private WNFS shares the same metadata structure as the [public WNFS](/spec/public-wnfs.md#metadata).

Or we just duplicate the link to the public WNFS metadata. I think it'd be useful here :)

spec/private-wnfs.md
```diff
@@ -112,57 +114,126 @@ type PrivateFile = {
...
+
+A viewing agent may be able to view more than a single revisions of a node. This information must be kept somewhere that some agents would be able to discover as they walk through a file system, but stay hidden from others. This is achieved per node with a "node key". Every revision of a node MUST have a unique skip ratchet, bare namefilter, and i-number.
+
+The skip ratchet is the single source of truth for generating the decryption key. Knowledge of this one internal skip ratchet state is sufficient to grant access to all of the relevant state in the diagram:
+* Generate the content key for the current node
+* Generate decryption pointers for future versions of this node
+* Access to decryption pointers to all child nodes
+* Access to the skip ratchets for all child nodes
+
+![](./diagrams/node_key_layout.svg)
```
matheus23, line 204: I'm thinking the "Temporal Header" here may be misleading:
It looks like that's a single, encrypted ciphertext, but what is described in the types are actually a ciphertext block for every key.

---

I know you mostly updated the previous diagram which was already slightly outdated. I think the "Private Content" box should encompass both boxes to the left of it, because its ciphertext contains the ciphertext of the boxes to the left. That's on me though, so I can also go ahead and fix it in this PR if you want or after this PR.

spec/private-wnfs.md
```diff
@@ -112,57 +114,126 @@ type PrivateFile = {
...
+
+The decrypted (cleartext) file layer is very straightforward: it follows the exact interface as public WNFS files and directories. The primary difference is that while the public file system MUST form a DAG by its hash-linked structure, special care MUST be taken so that the private file system does not form pointer cycles.
+
+# 4 Algorithms
+
+All algorithms MUST have access to a `PrivateForest` in their context.
+
+## 4.1 Namefilter Hash Resolution
+
+`resolveHashedKey: Hash<Namefilter> -> (Namefilter, Encrypted<PrivateNodeHeader>, Array<Encrypted<PrivateNode>>)`
```
matheus23, line 216:
```suggestion
`resolveHashedKey : Hash<Namefilter> -> (Namefilter, Encrypted<PrivateNodeHeader>, Array<Encrypted<PrivateNode>>)`
```

for consistency with the below?

Also, I forgot to update this type when I changed the private filesystem structure. It should probably just be simplifited to

```suggestion
`resolveHashedKey : Hash<Namefilter> -> (Namefilter, Array<CID>)`
```

spec/private-wnfs.md
```diff
@@ -172,34 +243,43 @@ Where
...
-- Seeking & repairing: Similar to seeking, this mode will look for the most recent version of a file and once it found it and in case it's not the same as what the parent refers to, it'll repair the parent's link to link to the most recent revision.
+### 4.3.1 Current Snapshot
 
-In all of these cases the next path segment's directory or file's hash of the namefilter can be retrieved by accessing the current directory's `directory.entries[segmentName].name`, looking the private node up as described in [Namefilter Hash Resolving](#Namefilter-Hash-Resolving) and then decrypting the content node(s) using `directory.entries[segmentName].contentKey`.
+Resolve the current snapshot: only resolve the current snapshot of a version. This only requires a content key for decryption.
+
+### 4.3.2 Seek
+
+For each path segment, look up the most recent version that can be found (as described in the [private versioning algorithm](#Private-Versioning)). This requires access to the directory's revision key.
```
matheus23, line 262:
```suggestion
For each path segment, look up the most recent version that can be found (as described in the [private versioning algorithm](#Private-Versioning)). This requires access to the directory's node key.
```

spec/private-wnfs.md
```diff
@@ -172,34 +243,43 @@ Where
...
+
+#### 4.3.2.1 Attach
+
+A variant of seeking. This mode searches for the latest revision of a node (by its namefilter and skip ratchet) and if it is found to differ from the parent's link, a new parent revision MAY be created with an updated link to the file. It is RECOMMENDED that this process then be performed recursively to the highest parent that the agent has write access to. This saves the next viewer from having to seek forward more than is strictly nessesary, as this always starts from the parent's link which moves forward monotonically.
+
+In all of these cases the next path segment's directory or file's hash of the namefilter MUST be retrieved by accessing the current directory's `directory.entries[segmentName].name`, looking up the private node as described in [Namefilter Hash Resolutions](#41-Namefilter-Hash-Resolution) and then decrypting the content node(s) using `directory.entries[segmentName].contentKey`.
 
 If this mode is seeking, the `directory.entries[segmentName].revisionKey` needs to be decrypted using the revision key for the current directory.
```
matheus23, line 270:
```suggestion
If this mode is seeking, the `directory.entries[segmentName].revisionKey` needs to be decrypted using the node key for the current directory.
```

spec/private-wnfs.md
```diff
@@ -112,57 +114,126 @@ type PrivateFile = {
...
+
+If the child is a `Node`, repeat the process of with the next nibble.
+
+If the child is a HAMT bucket of values, iterate that bucket to find one that has a namefilter that matches the hash of the namefilter. The associated values then contains the ciphertexts and the algorithm is done.
+
+## 4.2 Private Versioning
+
+`toVersioned : (Namefilter, RevisionKey) -> Namefilter`
+
+Every private file or directory implicitly links to the name (namefilter) of its next version. These implicit links can only be resolved when you have the revision key that allows you to decrypt the `PrivateNodeHeader`. Given a `PrivateNodeHeader` it is possible to construct namefilters for newer versions of this private file or directory by stepping the ratchet forward as far as you want to look ahead. Then, the new namefilter is:
```
matheus23, line 236:
```suggestion
Every private file or directory implicitly links to the name (namefilter) of its next version. These implicit links can only be resolved when you have the node key that allows you to decrypt the `PrivateNodeHeader`. Given a `PrivateNodeHeader` it is possible to construct namefilters for newer versions of this private file or directory by stepping the ratchet forward as far as you want to look ahead. Then, the new namefilter is:
```

spec/private-wnfs.md
expede: The way that we're using bucket in the code is also a bit confusing. Many other associative data structures use "buckets" as the leaf (or other container) where data is held.

I'm happy to switch this terminology to "entry" or perhaps "multientry"/"multivalue", but we should add a section that clarifies the way we're using the word "bucket" as well

spec/private-wnfs.md
expede: > but what is described in the types are actually a ciphertext block for every key.

Tell me more about this! Could that leak how many elements there are in the directory?

> because its ciphertext contains the ciphertext of the boxes to the left.

Which of the following do you mean?

1. They're in one contiguous block and the user needs to know where the length to break multiple ciphertexts on
2. They're doubly-encrypted (one inside the other)
3. The private & temporal headers are in cleartext inside the private content

spec/private-wnfs.md
expede: That space is just my Haskell fingers. I agree that consistency is the most important thing here. Which do you think folks would find clearest coming from a Rust and TS background?

spec/private-wnfs.md
expede: Why not a revision key? "Node" is a very general term.

expede:
Yeah the more I think about this, the more I want to move the encrypted data layer (Dark Forest) to its own spec. We're going to reuse it for a few things e.g. Dialog 🤔 It doesn't have to be in this PR, just tossing it out there for later

spec/private-wnfs.md
matheus23: Yes...
It actually used to be `revision key` everywhere, then you changed it to `node key`.
I just wanted to make it consistent. Changing it back to saying "revision key" everywhere is fine by me :)

spec/private-wnfs.md
matheus23: USERLAND it is :stuck_out_tongue:  :+1:

spec/private-wnfs.md
matheus23: > Tell me more about this! Could that leak how many elements there are in the directory?
It's always exactly one encrypted temporal node (revision?) key next to every content key, all of these wrapped in another layer of encryption.
That leaks exactly the same amount of information as the version in the diagram. The size of the overall ciphertext may give you a rough idea of what the size of the directory may be.

> Which of the following do you mean?
(2)

spec/private-wnfs.md
matheus23: I don't think there's the `:` as a `type-of` operator in the rust syntax.
Technically the `identifier: Type` syntax exists in typescript records.

I honestly could go either way. After all, I really like the Elm/Idris/Haskell style with an extra space. But maybe we just go for the more common format from typescript?

matheus23:
@expede This is really close!
I'd love to get the node key vs. revision key thing sorted out as well as the temporal header in the diagram.
Then we can merge this :raised_hands:

spec/private-wnfs.md
```diff
@@ -1,80 +1,94 @@
...
-  >>
+At the encrypted data layer, the private forest is a collection of ciphertext blocks. These blocks SHOULD be smaller than 256 kilobytes in order to comply with the default IPFS block size. Keeping block size small is also useful for reducing metadata leakage - it's less obvious what the file size distribution in the private file system is like if these files are split into blocks.
+
+Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multimap](https://en.wikipedia.org/wiki/Multimap). The HAMT MUST have a node-degree of 16[^1], and MUST used saturated  saturated [namefilter](/spec/namefilter.md)s as keys.
+
+### 2.1.1 Data Types
+
+The multimap container MUST be represented as a CBOR-encoded Merkle HAMT. The values MUST be a set of IPLD-formatted binary blobs.
 
+All values in the Merkle HAMT MUST be sorted in lexicographic ascending order by CID.
```
appcypher, line 42: If I'm correct, `lexicographic` here refers to a generalized form of alphabetical ordering for some ordered set? So we don't have to convert CIDs temporarily to strings to apply ordering. I'm asking because CIDs are represented internally as bytes and some metadata.

OTOH, the CID library we use determines order using information in a CID data structure.

- version
- codec
- digest size
- multihash code, and
- digest bytes

spec/private-wnfs.md
expede: haha well I guess it makes more sense once you dig in 😅

matheus23 (approved):

matheus23:
Thank you very much!
We can go ahead and merge now :)
Go go

PR #36, opened 2022-08-30
Suggest fixes

appcypher:


matheus23:
@appcypher why close this? There's some good suggestions in here.
The PR will get slightly stale once we merge #32, which I'd like to merge first, but let's just rebase once that landed :)

appcypher:
> The PR will get slightly stale once we merge https://github.com/wnfs-wg/spec/pull/32

That's what I thought. And since #32 is WIP, better to suggest changes there.

PR #39, opened 2022-09-23
Revise diagram about encryption components of private node

matheus23:


matheus23:
Still need to figure out how exactly to get this laid out so it looks nicer.
This is more like what I thought of. Every time I use a cluster in this diagram, it's intended to be one instance of a ciphertext.

expede:
@matheus23 as I read this diagram, you have several levels of nested encryption... is that right?

matheus23:
@expede Yes. This is what I raised in https://github.com/wnfs-wg/spec/pull/32#discussion_r957097694 originally, and this PR is actually the follow-up to that, that I mentioned.

The reason it's nested is basically outlined here: https://github.com/wnfs-wg/spec/pull/27#issuecomment-1206552240
We don't want to have two accumulators per-node, but we also don't want to have node ciphertexts being distinguishable from file shard ciphertexts.

Sorry I should've had a better PR description (any description would've been helpful! :sweat_smile:).

expede:
@matheus23 I think we may need to talk this one through live, then. The multiple layers of encryption seems "off", and (if I was understanding them correctly) the layout in the original diagrams seemed to have several advantages over the nested encryption. Let's find some time on Monday.

matheus23:
We came to the conclusion that we can and want to change this. I've captured this in #47.

I think we should still merge this PR as it makes the spec text consistent with the spec diagrams.

expede:
@matheus23 Works for me! Thank you for splitting that out into #47. Prioritization!!

matheus23:
Does that mean you'd accept this PR?

matheus23:
Closing in favor of #58

PR #40, opened 2022-09-23
Correct signature for hash resolution

matheus23:


spec/private-wnfs.md
```diff
@@ -226,9 +226,9 @@ All algorithms MUST have access to a `PrivateForest` in their context.
 
 ## 4.1 Namefilter Hash Resolution
 
-`resolveHashedKey: Hash<Namefilter> -> (Namefilter, Encrypted<PrivateNodeHeader>, Array<Encrypted<PrivateNode>>)`
+`resolveHashedKey: Hash<Namefilter> -> (Namefilter, Array<Encrypted<PrivateNode>>)`
 
-The private file system is a pointer machine, where pointers MUST be hashes of namefilters. To resolve a namefilter hash, look up the hash in the HAMT. The resulting key-value pair MUST contain the full "expanded" namefilter, the private node header, and a list of at least one private node.
+The private file system is a pointer machine, where pointers MUST be hashes of namefilters. To resolve a namefilter hash, look up the hash in the HAMT. The resulting key-value pair MUST contain the full "expanded" namefilter and a list of at least one private node.
```
expede, line 231: We don't need the header separated anymore? Or is this to keep the semantic layers separate?

matheus23: Basically the `PrivateNode` contains the encrypted `PrivateNodeHeader`, so we don't need to return it additionally.

This is an artifact from a previous change: https://github.com/wnfs-wg/spec/pull/27#issuecomment-1206552240

expede (approved):

PR #41, opened 2022-09-23
Add private backlinks

### LINKED ISSUE #34: Private Merge: Identify history "tracks" for files/directories
# The Problem

When two private file systems get merged, you end up with multiple nodes at each revision, for example:
1. Bob writes a version of "file.txt": "Hello "
2. Alice writes two versions of "file.txt" on top of that: + "Alice", then + "!"
3. Bob writes two versions of "file.txt" concurrently on top of (1): + "Bob", then + "!"

After merging, you end up with the revisions looking like this:
| Revision | Content |
|-:|-|
| 1 | "Hello " |
| 2 | "Hello Alice", "Hello Bob" |
| 3 | "Hello Bob!", "Hello Alice!" |

Notice that the order may get "mixed up", because the order is determined by something independent of the contents (the ciphertext's hash).

When looking at the "Hello Bob!" file.txt version of revision 3, it's not clear what *that* particular version's predecessor is. We only know its previous version is included in the set of "Hello Alice" and "Hello Bob".
Being able to recover the history DAG of a file/directory should be possible from the data model for as far as you have access to read.

# Possible Solution(s)

We introduce a previous pointer array similar to what exists in the public file system. This previous pointer should identify which previous ciphertext was the previous version.
If the previous pointer is the CID of the previous ciphertext this forces the causal dependency to be possible (you can't "guess" the hash of something that you don't know about).
However, this leaks metadata: Even though a client can't *read* the previous version, they can see which of the ciphertexts was the previous one.
We *could* decide that that's very little metadata and we shouldn't worry about that, but could also do a little better by encrypting the previous pointer with the *previous* revision's node (ratchet) key.
That way, only clients that can decrypt the previous revision can read the backpointers for the future revision.
A writer can obviously encrypt these backpointers because they should have access to read revision N in order to write revision N+1.
The size of the encrypted backpointers array will likely still be leaked, but I think that's not worth compromising efficiency for. The "danger" here is that someone who can *read the current revision* knows *how many previous versions* this file was based on. I.e. they know the "N" in the n-ary merge, they know that this file was just created (no backpointers) or the file had a single previous version.

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/private-backlinks/spec/private-wnfs.md#31-cleartext-data)

Closes #34

### DIFF
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -149,9 +149,9 @@ Node headers MUST be encrypted with the key derived from the node's skip ratchet
 
 Node metadata is the userland equivalent of the node's header.
 
-### 3.1.3 Node Previous
+### 3.1.3 Node `previous` Backlinks
 
-Node previous is the encrypted binary ascending order sorted CBOR list of CIDs this node was based on when constructed. This value MUST be encrypted with the ratchet key derived from the previous skip ratchet relative to the current node.
+The node "previous" links is an encrypted binary ascending order sorted CBOR list of CIDs this node was based on when constructed. This value MUST be encrypted with the ratchet key derived from the previous skip ratchet relative to the current node.
 Note that the array of previous CIDs can refer to private nodes that are 'older' than one revision.
 
 ### 3.1.4 Private File
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -153,8 +153,9 @@ Node metadata is the userland equivalent of the node's header.
 
 The `previous` link provides an encrypted back-pointer to any direct causal dependencies (the previous version of this node). The encryption protects from metadata leakage from agents that do not have access to the relevant node.
 
-The `previous` link MUST be a encrypted CBOR list sorted by CID in binary ascending order. This value MUST be encrypted with the same skip ratchet key as referred to in the particular link.
-Note that the array of previous CIDs can refer to private nodes that are 'older' than one revision.
+The `previous` link MUST be an encrypted CBOR list of CIDs sorted in binary ascending order. This value MUST be encrypted with the same skip ratchet key as referred to in the particular link.
+
+Each CID in the decrypted `previous` links MUST refer to a value from the private forest. At least once CID MUST refer to a CID from the previous revision. If the `previous` links contain more than one element, then some CIDs MAY refer to CIDs of even older revisions. `previous` link CIDs MUST NOT refer to values in the private forest from newer revisions.
 
 ### 3.1.4 Private File
 
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -155,7 +155,9 @@ The `previous` link provides an encrypted back-pointer to any direct causal depe
 
 The `previous` link MUST be an encrypted CBOR list of CIDs sorted in binary ascending order. This value MUST be encrypted with the same skip ratchet key as referred to in the particular link.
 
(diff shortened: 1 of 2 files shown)


spec/private-wnfs.md
```diff
@@ -139,47 +141,52 @@ type ExternalContent = {
...
 
 Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to [§3.2.3 Pointers & Keys](#323-pointers--keys) for more detail.
 
-### 3.2.2 Node Metadata
+### 3.1.2 Node Metadata
 
 Node metadata is the userland equivalent of the node's header.
 
-### 3.2.1 Private File
+### 3.1.3 Node Previous
```
expede, line 152: Friendly suggestion: perhaps "previous link" or "node previous link" (or wrap it in scare quotes). I had to read the sentence below a couple times to realize that it wasn't a grammatical error and you were trying to say "the previous link on this node"

matheus23: Yeah, I agree. I changed it to "Node `previous` Backlinks" and reworded the first sentence a bit.

spec/private-wnfs.md
```diff
@@ -204,7 +211,7 @@ In the above diagram, newer revisions of nodes progress left-to-right. The file
 
 Special attention should be paid to the relationship of the skip ratchet to content keys. There is a parallel structure between the skip ratchets and the content hierarchy. The primary difference is that access to _only_ a content key does not grant access to other revisions, where having access to a skip ratchet includes the next revisions and the ability to derive the content key.
 
-### 3.2.4.2 Revision Key Structure
+### 3.1.7.2 Revision Key Structure
```
expede, line 214: We really need an auto-number-er of some kind 🤔

spec/private-wnfs.md
```diff
@@ -110,6 +110,8 @@ type PrivateDirectory = {
   version: "0.2.0"
   // encrypted using deriveKey(ratchet)
   header: Encrypted<CBOR<PrivateNodeHeader>>
+  // encrypted using deriveKey(previousRatchet) where inc(previousRatchet) = ratchet
+  previous: Encrypted<CBOR<Array<CID>>>
```
expede, line 114: Not sure if good idea or bad idea: `s/previous/history/` 🤔

expede: No, definitely not. "History" implies that it's the entire history. I think we had this discussion before 😛

spec/private-wnfs.md
```diff
@@ -139,47 +141,52 @@ type ExternalContent = {
...
 
-### 3.2.2 Node Metadata
+### 3.1.2 Node Metadata
 
 Node metadata is the userland equivalent of the node's header.
 
-### 3.2.1 Private File
+### 3.1.3 Node `previous` Backlinks
+
+The node "previous" links is an encrypted binary ascending order sorted CBOR list of CIDs this node was based on when constructed. This value MUST be encrypted with the ratchet key derived from the previous skip ratchet relative to the current node.
```
expede, line 154: Style suggestion:

```suggestion
The `previous` link provides an encrypted back-pointer to any direct causal dependencies (the previous version of this node). The encryption protects from metadata leakage from agents that do not have access to the relevant node.

The `previous` link MUST be a encrypted CBOR list sorted by CID in binary ascending order. This value MUST be encrypted with the same skip ratchet key as referred to in the particular link.
```

(The above could probably be tighter, but you get the idea)

expede (approved):
Approved, but with some suggestions on wording

spec/private-wnfs.md
matheus23: Agree, but also, why do we number these sections?

Issue #10, opened 2022-06-27
Private File Sharding


matheus23:
What the whitepaper says about file sharding at the moment boils down to "store a ratchet for the key and the amount of segments" plus this:
![image](https://user-images.githubusercontent.com/1430958/176010392-6b0380fa-6dbe-4034-bfb8-591a0fa28037.png)

I think we can improve upon this a little bit:
1. I'd call into question whether we really need a whole skip ratchet for the file contents. Why not just a single key? What a skip ratchet gives you is the ability to give someone read access to all revisions of certain file *contents* without giving them access to any file *metadata*. How often do you really want to share all file contents without giving them access to the metadata?
2. For every version we still have to re-encrypt the whole file, even if only a single byte was inserted. We could do some cypherchunking scheme instead.

PR #44, opened 2022-10-19
Private File Sharding

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/file-sharding/spec/private-wnfs.md)

@expede: I don't know what the purpose of the skip ratchet was in private file sharding, so I'm proposing this way of sharding files without a skip ratchet in between.

Design goals were:
- every block has a limited size
- avoid having a metadata change in the PrivateFile header requiring a re-encrypt of any content blocks (having the file content encryption keys be derived from the header's skip ratchet would have meant that. So we avoid that by keeping another key around that is allowed to stay the same across revisions)
- you can seek efficiently in the file content (this is now O(1))
- make it possible to share access to only a file's content at a single revision without sharing access to all future revisions (to do this, just share the `key` and bare namefilter)

Design non-goals:
- efficient inserts/removes/in-place updates in the file content (The UnixFS file protocol supports that. In practice I don't know *anyone* making use of that. It's considerably harder to implement & provide interfaces for. Could be a future-future thing) 
- make it possible to share read access to only a file's content, while giving access to all revisions temporally (this *could* in theory be enabled by making `key` a skip ratchet instead of randomly-generated every revision. I don't think that's particularily useful though and may not be worth the complexity)

Work on #10

### DIFF
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -139,6 +139,7 @@ type PrivateFile = {
 
 type ExternalContent = {
   key: Key
+  blockSize: Uint64 // in bytes, at max 262132
   blockCount: Uint64
 }
 ```
@@ -171,13 +172,17 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-shared-file-content-access) for more detail.
 
-The block count for externalized content MUST reference the number of blocks the externalized content was split into.
+The block size MUST be at least 1 and at maximum $2^{18} - 12 = 262,322$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes need to be prepended to each ciphertext. It is RECOMMENDED to use the maximum block size.
+
+The block count MUST reference the number of blocks the externalized content was split into.
 
 The externalized content's `key` MUST be regenerated randomly whenever the file content changes. If the content stays the same across metadata changes, the content key MAY remain the same across those revisions
 
 NB: Label namefilters MUST be computed as described in the algorithm for [sharded file content access](#44-sharded-file-content-access).
 
-Each multi-value in the HAMT MUST have exactly one CID. That CID refers to a ciphertext block. Each block MUST be a $2^{18} = 262144$ byte ciphertext, where the first 12 bytes are the initialization vector for encryption, so each block MUST encode exactly 262,132 plaintext bytes, except for the last block with index equal to `blockCount - 1`, which MAY be smaller, but MUST NOT encode an empty plaintext.
+Entries in the private forest corresponding to externalized content blocks MUST have exactly one CID as their multi-value. This CID MUST refer to a ciphertext with exactly `12 + blockSize` bytes, except for the last block with index `blockCount - 1`. The first 12 bytes refer to the initialization vector and the rest are the ciphertext.
+
+If any externalized content blocks exceed the specified `blockSize` or are missing in the private forest despite having a lower index than `blockCount` during file read operations, then these operations MUST produce an error.
 
 ### 3.1.5 Private Directory
 
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -112,6 +112,8 @@ type PrivateDirectory = {
   version: "0.2.0"
   // encrypted using deriveKey(ratchet)
   header: Encrypted<CBOR<PrivateNodeHeader>>
+  // encrypted using deriveKey(previousRatchet) where inc(previousRatchet) = ratchet
+  previous: Encrypted<CBOR<Array<CID>>>
 
   // USERLAND
   metadata: Metadata
@@ -141,21 +143,31 @@ type ExternalContent = {
 }
 ```
 
-### 3.2.1 Node Headers
+### 3.1.1 Node Headers
 
 Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to the section [Pointers & Keys](#323-pointers--keys) for more detail.
 
-### 3.2.2 Node Metadata
+### 3.1.2 Node Metadata
 
 Node metadata is the userland equivalent of the node's header.
 
-### 3.2.1 Private File
+### 3.1.3 Node `previous` Backlinks
+
+The `previous` link provides an encrypted back-pointer to any direct causal dependencies (the previous version of this node). The encryption protects from metadata leakage from agents that do not have access to the relevant node.
+
+The `previous` link MUST be an encrypted CBOR list of CIDs sorted in binary ascending order. This value MUST be encrypted with the same skip ratchet key as referred to in the particular link.
+
+Each CID in the decrypted `previous` links MUST refer to a value from the private forest. Either the list of CIDs is empty or at least one CID MUST refer to a CID from the previous revision.
+
+If the `previous` links contain more than one element, then some CIDs MAY refer to CIDs of even older revisions. `previous` link CIDs MUST NOT refer to values in the private forest from newer revisions.
+
+### 3.1.4 Private File
 
 A private file MUST contain the actual bytes that represent the file. Files MAY also contain userland metadata.
 
 Private file content has two variants: inlined or externalized. Externalized content is held as a separate node in the bucket. Inlined content is kept alongside (and thus is decrypted with) the header.
 
-#### 3.2.1.1 Externalized Content
+#### 3.1.4.1 Externalized Content
 
 Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-shared-file-content-access) for more detail.
 
@@ -167,27 +179,27 @@ NB: Label namefilters MUST be computed as described in the algorithm for [sharde
 
 Each multi-value in the HAMT MUST have exactly one CID. That CID refers to a ciphertext block. Each block MUST be a $2^{18} = 262144$ byte ciphertext, where the first 12 bytes are the initialization vector for encryption, so each block MUST encode exactly 262,132 plaintext bytes, except for the last block with index equal to `blockCount - 1`, which MAY be smaller, but MUST NOT encode an empty plaintext.
 
-### 3.2.1 Private Directory
+### 3.1.5 Private Directory
 
 A private directory MUST contain links to zero or more further nodes. Private directories MAY include userland metadata.
 
 See the section for [Read Hierarchy](#324-read-hierarchy) for more information about the link structure.
 
-### 3.2.3 Pointers & Keys
+### 3.1.6 Pointers & Keys
 
 Keys are always attached to pointers to some data.
 
 ![](./diagrams/decryption_pointer.svg)
 
-#### 3.2.3.1 Revision Key
+#### 3.1.6.1 Revision Key
 
 revision keys MUST be derived from the skip ratchet for that node, incremented to the relevant revision number. This limits the reader to reading from a their earliest ratchet and forward, but never earlier revisions than that.
 
-#### 3.2.3.2 Content Key
+#### 3.1.6.2 Content Key
 
 Content keys MUST be derived from the [Revision Key](#3231-revision-key) by hashing it with SHA3. The content key grants access to a single revision snapshot of that node and its children, but no other revisions forward or backward.
 
-### 3.2.4 Read Hierarchy
+### 3.1.7 Read Hierarchy
 
 Access in WNFS is fundamentally hierarchical. Access granted to a single node in a DAG implies access to all of its child nodes (and no others). Decryption pointers provide a way to "discover" the structure of the portion of the file system accessible to the viewer. This process is always started from a pointer held by the viewer outside of the file system.
 
@@ -199,7 +211,7 @@ Each link in the above picture is looked up in the [encrypted HAMT](#211-data-ty
 
 Note that holding the decryption pointer to this particular directory MUST NOT grant access to sibling or parent nodes, nor to other structures rooted in the private forest.
 
-### 3.2.4.1 Temporal Hierarchy
+### 3.1.7.1 Temporal Hierarchy
 
 Being a versioned file system, private nodes also have read control in the temporal dimension as well as in the [file read hierarchy](#324-read-hierarchy). An agent MAY have access to one or more revisions of a node, and the associated children in that temporal window.
 
@@ -212,7 +224,7 @@ In the above diagram, newer revisions of nodes progress left-to-right. The file 
 
 Special attention should be paid to the relationship of the skip ratchet to content keys. There is a parallel structure between the skip ratchets and the content hierarchy. The primary difference is that access to _only_ a content key does not grant access to other revisions, where having access to a skip ratchet includes the next revisions and the ability to derive the content key.
 
-### 3.2.4.2 Revision Key Structure
+### 3.1.7.2 Revision Key Structure
 
 A viewing agent may be able to view more than a single revisions of a node. This information must be kept somewhere that some agents would be able to discover as they walk through a file system, but stay hidden from others. This is achieved per node with a "revision key". Every revision of a node MUST have a unique skip ratchet, bare namefilter, and i-number.
 
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -161,6 +161,8 @@ The block count for externalized content MUST reference the number of blocks the
 
 Each block MUST be a $2^{18} = 262144$ byte ciphertext, where the first 12 bytes are the initialization vector for encryption, so each block MUST encode exactly 262,132 plaintext bytes, except for the last block with index equal to `blockCount - 1`, which MAY be smaller.
 
+The externalized content's `key` MUST be re-generated randomly every time the file content changes. It MAY stay the same across private file header revisions if only file metadata changes.
+
 ### 3.2.1 Private Directory
 
 A private directory MUST contain links to zero or more further nodes. Private directories MAY include userland metadata.
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -74,7 +74,11 @@ A space optimization delaying the creation of additional layers until 3 collisio
 
 #### 2.1.1.3 `Entry`
 
-A leaf node, containing the expanded label as well as a value. As the private forest for WNFS, this MUST contain a namefilter as label and a set of CIDs of ciphertexts of conflicting writes as the value.
+An `Entry` node MUST consist of:
+* The expanded label (HAMT hash preimage)
+* The value for this label
+
+If the HAMT is used as the `PrivateForest` for WNFS, then the values stored SHOULD be ciphertexts representing conflicting file system writes to that same path and revision.
 
 ## 2.2 Ciphertext Files
 
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -6,7 +6,7 @@ The private file system provides granular control over read access along two dim
 
 # 1 Terminology
 
-Encryption adds another dimension to a file system: visibility. The data and file layers are each augmented with cleartext and ciphertext components. While namefilters and multivalues do encode a concept of an encrypted file, we generally only speak of the data layer. 
+Encryption adds another dimension to a file system: visibility. The data and file layers are each augmented with cleartext and ciphertext components. While namefilters and multivalues do encode a concept of an encrypted file, we generally only speak of the data layer.
 
 ![](./diagrams/layer_dimensions.svg)
 
@@ -15,7 +15,7 @@ Broadly speaking, there is a "decrypted" layer and an "encrypted" layer.
 - The "decrypted" layer defines the type of data you can decrypt given the correct keys. Links between blocks in this layer are references in the HAMT data structure at the "encrypted" layer.
 - The "encrypted" layer defines how all of the encrypted data blocks are organized as IPLD data. Links in this layer are CID-links.
 
-These all form graphs, where the nodes and links have different meanings per layer. The 
+These all form graphs, where the nodes and links have different meanings per layer.
 
 | Visibility | Layer | Node        | Link             |
 |------------|-------|-------------|------------------|
@@ -43,7 +43,7 @@ All values in the Merkle HAMT MUST be sorted in binary ascending order by CID.
 
 ```typescript
 type PrivateForest = CBOR<HAMT<Namefilter, Array<CID<ByteArray>>>>
-  
+
 type HAMT<K, V> = {
   structure: "hamt"
   version: "0.1.0"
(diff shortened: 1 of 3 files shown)


spec/private-wnfs.md
```diff
@@ -155,9 +157,15 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 #### 3.2.1.1 Externalized Content
 
-Since external content is separate from the header, it MUST have a unique namefilter derived from a ratchet (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [the relevant algorithm](#44-shared-file-content-access) for more detail.
+Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [the relevant algorithm](#44-shared-file-content-access) for more detail.
+
+The block count for externalized content MUST reference the number of blocks the externalized content was split into.
```
expede, line 162: What if it's wrong? i.e. the block claims that there's 100 blocks, but there's only 5

spec/private-wnfs.md
```diff
@@ -155,9 +157,15 @@ Private file content has two variants: inlined or externalized. Externalized con
...
+
+The externalized content's `key` MUST be re-generated randomly every time the file content changes. It MAY stay the same across private file header revisions if only file metadata changes.
```
expede, line 164:
```suggestion
The externalized content's `key` MUST be regenerated randomly whenever the file content changes. If the content stays the same across metadata changes, the content key MAY remain the same across those revisions
```

expede: Will the content sit under both labels, then? If so, can/should we at least change the CID somehow? 🤔 (e.g. prefix with a pepper)

spec/private-wnfs.md
```diff
@@ -155,9 +157,15 @@ Private file content has two variants: inlined or externalized. Externalized con
...
 
-The skip ratchet counter for externalized content MUST reference the block index, not the version. The skip ratchet for the externalized blocks is generated fresh per revision, where the header maintains the temporal access. For structural sharing between revisions, decryption pointers ("flattened" derived keys) MUST be used in order to not leak non-shared chunks on that same skip ratchet.
+The namefilters to be used as labels for the ciphertexts in the HAMT are computed as defined in [Algorithm 4.4: Sharded File Content Access](#44-sharded-file-content-access).
```
expede, line 166:
```suggestion
NB: label namefilters MUST be computed per [Algorithm 4.4](#44-sharded-file-content-access)
```

spec/private-wnfs.md
```diff
@@ -155,9 +157,15 @@ Private file content has two variants: inlined or externalized. Externalized con
...
+
+Each multi-value in the HAMT MUST have exactly one CID. That CID refers to a ciphertext block. Each block MUST be a $2^{18} = 262144$ byte ciphertext, where the first 12 bytes are the initialization vector for encryption, so each block MUST encode exactly 262,132 plaintext bytes, except for the last block with index equal to `blockCount - 1`, which MAY be smaller, but MUST NOT encode an empty plaintext.
```
expede, line 168:
```suggestion
Each HAMT value MUST be referenced by exactly one CID. That CID refers to a ciphertext block. Ciphertexts MUST NOT be black (MUST have at least one byte). A ciphertext chunk MUST NOT exceed $2^{18} = 262144$ bytes of ciphertext, plus a 12-byte initialization vector yielding exactly $262,132$ plaintext bytes.
```

### Chunk Size

I think lifting the forced chunk size may be a good thing, since it would allow us to break chunks on e.g. video frame boundaries.

Determinism would be nice in the general case, but is possibly not _required_ per se. I suggest that we RECOMMEND $2^{18}$ bytes, but let devs lay their data out as needed.

Thoughts?

### Multivalue

I think what you're trying to say is that the each value in the multivalue forms a tree? If so, it's not totally clear from the text as it stands, or in my quick copy edit suggestion above.

bar

If so, have you considered making them totally flat, and letting the user derive the entry point via index number? Something like this:

bar

Where the hash is derived from the index number plus the CID of the first chunk...

$$
\textsf{hash}(index \|| firstCID \|| aesKey)
$$

...or something?

spec/private-wnfs.md
```diff
@@ -272,7 +280,7 @@ Resolve the current snapshot: only resolve the current snapshot of a version. Th
 
 ### 4.3.2 Seek
 
-For each path segment, look up the most recent version that can be found (as described in the [private versioning algorithm](#Private-Versioning)). This requires access to the directory's revision key.
+For each path segment, look up the most recent version that can be found (as described in the [private versioning algorithm](#42-Private-Versioning)). This requires access to the directory's revision key.
```
expede, line 283: Nit: an earlier instance referenced it by algorithm number, but this is by lowercase name. I personally like this one more for a few reasons and would change the other one:

* Reads more "naturally"
* You don't have to update all of the section numbers in the links (it's a huge PITA)

spec/private-wnfs.md
```diff
@@ -284,33 +292,31 @@ If this mode is seeking, the `directory.entries[segmentName].revisionKey` needs
 
 ##### 4.3.2.1.1 Example
 
-To illustrate this mode, consider the following diagram. An agent may only have access to some nodes, but not their parent. The agent is able to create new revisions of files and directories, and link that back to the previous version. Some number of revisions may accrue before this can be fully rooted again. Attachment occurs when a reader with enough rights to perform the attachment inspects the file system and discovers that they can attach this file to its parents. 
+To illustrate this mode, consider the following diagram. An agent may only have access to some nodes, but not their parent. The agent is able to create new revisions of files and directories, and link that back to the previous version. Some number of revisions may accrue before this can be fully rooted again. Attachment occurs when a reader with enough rights to perform the attachment inspects the file system and discovers that they can attach this file to its parents.
```
expede, line 295: (Wordy)

```suggestion
Consider the following diagram. An agent may only have access to some nodes, but not their parent. The agent is able to create new revisions of files and directories, and link that back to the previous version. Some number of revisions may accrue before this can be fully rooted again. Attachment occurs when a reader with enough rights to perform the attachment inspects the file system and discovers that they can attach this file to its parents.
```

expede (changes requested):

spec/private-wnfs.md
```diff
@@ -37,30 +37,32 @@ Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multi
...
 
-All values in the Merkle HAMT MUST be sorted in binary ascending order by CID.
+All values in the Merkle HAMT MUST be sorted in binary ascending order by CID and MUST NOT contain duplicates.
 
 ```typescript
 type PrivateForest = CBOR<HAMT<Namefilter, Array<CID<ByteArray>>>>
-  
-type HAMT<K, V> = {
+
+type HAMT<L, V> = {
```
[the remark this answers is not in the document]
matheus23: Haha. Yeah Rust did. But this is typescript, and even in typescript this is 'normal'. (Just like in Java, etc.)

I *think* I'd prefer it this way for clarity. The `L` is supposed to mean "label", but it should also be evident that it *could* be something else.
I liked the fact that we can describe the private forest as `type PrivateForest = HAMT<Namefilter, Array<CID>>`, for anyone who knows about HAMTs already.
But we could also just remove all the abstraction and generics and make it specific to WNFS.

spec/private-wnfs.md
```diff
@@ -37,30 +37,32 @@ Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multi
 
 ### 2.1.1 Data Types
 
-The multimap container MUST be represented as a CBOR-encoded Merkle HAMT. The values MUST be a set of IPLD-formatted binary blobs.
+The multimap container MUST be represented as a CBOR-encoded Merkle HAMT. The values MUST be a set of `raw` codec CIDs.
```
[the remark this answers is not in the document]
matheus23: Yeah good point!

spec/private-wnfs.md
```diff
@@ -37,30 +37,32 @@ Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multi
...
 
-All values in the Merkle HAMT MUST be sorted in binary ascending order by CID.
+All values in the Merkle HAMT MUST be sorted in binary ascending order by CID and MUST NOT contain duplicates.
```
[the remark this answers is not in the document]
matheus23: Yep. It's not dangerous. In fact `rs-wnfs` would handle that gracefully, I'm pretty sure. It'd just consume the vector and deduplicate it when ingesting it into a `BTreeSet`.

spec/private-wnfs.md
```diff
@@ -37,30 +37,32 @@ Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multi
...
-  root: SparseNode<K, V>
+  root: SparseNode<L, V>
 }
 
-type SparseNode<K, V> = [
+type SparseNode<L, V> = [
   ByteArray<2>, // Sparse Index
-  Array<Entry<K, V>> // Entries
+  Array<Entry<L, V>> // Entries
 ]
```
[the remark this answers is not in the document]
matheus23: This... is supposed to be pseudo-typescript :sweat_smile:
I describe it a little bit in https://github.com/wnfs-wg/spec/blob/main/spec/notation.md.

The spec started out as IPLD schemas. Unfortunately they're not expressive enough for my purposes: All private data would just be `Bytes`. But I want to describe the data that's *inside*. I also *need* to describe the *actual* codec used (dag-cbor) for data that's encrypted.
Right now I manage via the special notation, e.g. `Encrypted<>` and `CBOR<>`, plus comments, but these things just don't exist in IPLD schema land. I've also had some trouble even finding out what the *correct*/most up-to-date IPLD schema language syntax is.

I'd be down for *adding* IPLD schemas to the spec at some point, especially for public WNFS. But I'd like to keep the annotated pseudo-typescript around so we have something that describes how the encryption nests / how things are encoded / what is encrypted with which key.

spec/private-wnfs.md
```diff
@@ -72,7 +74,7 @@ A space optimization delaying the creation of additional layers until 3 collisio
 
 #### 2.1.1.3 `Entry`
 
-A multi-valued leaf node, containing a file data and write conflicts.
+A leaf node, containing the expanded label as well as a value. As the private forest for WNFS, this MUST contain a namefilter as label and a set of CIDs of ciphertexts of conflicting writes as the value. This set of CIDs is also called the multi-value.
```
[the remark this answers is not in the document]
matheus23: Yep, I'm aware.
Just trying to introduce terms here. Is that something we should just assume the reader knows? I can remove that section.

spec/private-wnfs.md
matheus23: Then at some point reading may fail. Not much we can do about that, right?

We could technically just leave out the count :thinking:, as that's "gossipy", as Brendan likes to say. The end of the file would be reached in case the next block is missing / the first block that's not "full" is hit.

However, what would then happen if a single block is missing in the middle of a file? A client that wants to seek some bytes at the end of the file may read them just fine, but a client that reads from front to back would think it stops in the middle.

Having this gossipy count will let the client reading from front to back know that something went wrong and the file doesn't actually just stop in the middle.

spec/private-wnfs.md
matheus23: There is only a single label for both revisions in this case. The shard labels only depend on the file node's bare namefilter and `key` for for externalized content, (see [Algorithm 4.4](https://github.com/wnfs-wg/spec/blob/matheus23/file-sharding/spec/private-wnfs.md#44-sharded-file-content-access)).
So if the key stays the same across metadata changes, then all shard labels stay the same, too.

spec/private-wnfs.md
matheus23: > I think what you're trying to say is that the each value in the multivalue forms a tree?

Ah I guess the wording is pretty bad.
What I was trying to say is that for private file shards, each shard has its own namefilter, and in the HAMT, private shard namefilters must not be multi-value, but have a single value in the multi-value array.

---

> I think lifting the forced chunk size may be a good thing, since it would allow us to break chunks on e.g. video frame boundaries.

Ah, so essentially having a `blockSize` next to `blockCount` in the private file node? That sounds like a great idea. Then the 262,132 bytes just becomes the upper bound on that.

spec/private-wnfs.md
matheus23: I added some paragraphs about what should happen when it's wrong. Basically file read operations should error out as soon as they notice something is wrong.

spec/private-wnfs.md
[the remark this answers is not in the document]
expede: > I also need to describe the actual codec used (dag-cbor) for data that's encrypted.

Hmm I definitely see what you're trying to do, but I wonder if we could actually separate these concerns 🤔 I've struggled with the same in the past, but I wonder if instead of handling these at the same layer, we do really clear definitions of the cleartext and ciphertext layers, and show the translation between them via a function call. This means that the spec gets longer, but possibly more clear / easier to follow.

Let's chat about it on our weekly call :)

spec/private-wnfs.md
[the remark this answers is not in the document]
expede: Perhaps remove the paragraph and add a link to [multimaps](https://en.wikipedia.org/wiki/Multimap) or similar?

spec/private-wnfs.md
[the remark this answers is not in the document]
matheus23: In any case - I'm marking this as resolved, as I don't think this has to land in this PR :v:

spec/private-wnfs.md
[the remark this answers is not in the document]
matheus23: I've changed the wording to "This makes the private forest a mutlimap" with the link to wikipedia.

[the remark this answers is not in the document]
matheus23: Actually - I just removed it.

spec/private-wnfs.md
```diff
@@ -72,7 +74,7 @@ A space optimization delaying the creation of additional layers until 3 collisio
 
 #### 2.1.1.3 `Entry`
 
-A multi-valued leaf node, containing a file data and write conflicts.
+A leaf node, containing the expanded label as well as a value. As the private forest for WNFS, this MUST contain a namefilter as label and a set of CIDs of ciphertexts of conflicting writes as the value.
```
expede, line 77: Totally a nit, but a suggestion for consistent style:

```suggestion
An `Entry` node MUST contain the following children:

* The expanded label (HAMT hash preimage)
* The one or more CIDs of "values"

The values stored SHOULD be ciphertexts representing conflicting file system writes to that same path and revision.
```

spec/private-wnfs.md
```diff
@@ -136,14 +138,15 @@ type PrivateFile = {
...
+  key: Key
+  blockSize: Uint64 // in bytes, at max 262132
+  blockCount: Uint64
 }
 ```
 
 ### 3.1.1 Node Headers
 
-Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to [§3.2.3 Pointers & Keys](#323-pointers--keys) for more detail.
+Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to the section [Pointers & Keys](#323-pointers--keys) for more detail.
```
expede, line 149: Totally fine as-is, but it also reads well without the words `the section`:

```suggestion
Node headers MUST be encrypted with the key derived from the node's skip ratchet: the "content key". Headers MUST NOT grant access to other versions of the associated node. Node headers are in kernel space and MUST NOT be user writable. Refer to [Pointers & Keys](#323-pointers--keys) for more detail
```

spec/private-wnfs.md
```diff
@@ -167,15 +170,25 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 #### 3.1.4.1 Externalized Content
 
-Since external content is separate from the header, it MUST have a unique namefilter derived from a ratchet (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [the relevant algorithm](#44-shared-file-content-access) for more detail.
+Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-shared-file-content-access) for more detail.
+
+The block size MUST be at least 1 and at maximum $2^{18} - 12 = 262,322$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes need to be prepended to each ciphertext. It is RECOMMENDED to use the maximum block size.
+
+The block count MUST reference the number of blocks the externalized content was split into.
```
expede, line 177: What happens if this count is wrong? It this counter here for performance or is it integral to the lookup?

spec/private-wnfs.md
```diff
@@ -167,15 +170,25 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 #### 3.1.4.1 Externalized Content
 
-Since external content is separate from the header, it MUST have a unique namefilter derived from a ratchet (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [the relevant algorithm](#44-shared-file-content-access) for more detail.
+Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-shared-file-content-access) for more detail.
+
+The block size MUST be at least 1 and at maximum $2^{18} - 12 = 262,322$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes need to be prepended to each ciphertext. It is RECOMMENDED to use the maximum block size.
```
expede, line 175: Copyediting: run-on sentence

```suggestion
The block MUST NOT be empty, and SHOULD be no larger than $2^{18} - 12 = 262,322$ bytes. This allows both the 12-byte initialization vector and content to fit in the maximum IPFS block size.

Implementations are RECOMMENDED to use the maximum IPLD block size.
```

spec/private-wnfs.md
```diff
@@ -167,15 +170,25 @@ Private file content has two variants: inlined or externalized. Externalized con
...
+
+The block count MUST reference the number of blocks the externalized content was split into.
+
+The externalized content's `key` MUST be regenerated randomly whenever the file content changes. If the content stays the same across metadata changes, the content key MAY remain the same across those revisions
```
expede, line 179: 👍

spec/private-wnfs.md
```diff
@@ -167,15 +170,25 @@ Private file content has two variants: inlined or externalized. Externalized con
...
 
-The skip ratchet counter for externalized content MUST reference the block index, not the version. The skip ratchet for the externalized blocks is generated fresh per revision, where the header maintains the temporal access. For structural sharing between revisions, decryption pointers ("flattened" derived keys) MUST be used in order to not leak non-shared chunks on that same skip ratchet.
+NB: Label namefilters MUST be computed as described in the algorithm for [sharded file content access](#44-sharded-file-content-access).
+
+Entries in the private forest corresponding to externalized content blocks MUST have exactly one CID as their multi-value. This CID MUST refer to a ciphertext with exactly `12 + blockSize` bytes, except for the last block with index `blockCount - 1`. The first 12 bytes refer to the initialization vector and the rest are the ciphertext.
```
expede, line 183: I won't comment on all of them, but stylistically:

```suggestion
Entries in the private forest corresponding to externalized content blocks MUST have exactly one CID as their multi-value. This CID MUST refer to a ciphertext with exactly `12 + blockSize` bytes, except for the last block with index `blockCount - 1`. The first 12 bytes of the block MUST be an initialization vector, and the rest MUST be the ciphertext.
```

spec/private-wnfs.md
```diff
@@ -167,15 +170,25 @@ Private file content has two variants: inlined or externalized. Externalized con
...
+
+If any externalized content blocks exceed the specified `blockSize` or are missing in the private forest despite having a lower index than `blockCount` during file read operations, then these operations MUST produce an error.
```
expede, line 185: Hmmm, we could consider specifying a minimum number of exceptions for implementations. The exact representation will be different, but we know that this exception is some kind of `BlockSizeExceeded`, exception for example 🤔

spec/private-wnfs.md
```diff
@@ -296,33 +309,31 @@ If this mode is seeking, the `directory.entries[segmentName].revisionKey` needs
...
+Consider the following diagram. An agent may only have access to some nodes, but not their parent. The agent is able to create new revisions of files and directories, and link that back to the previous version. Some number of revisions may accrue before this can be fully rooted again. Attachment occurs when a reader with enough rights to perform the attachment inspects the file system and discovers that they can attach this file to its parents.
 
 ![](./diagrams/attach.svg)
 
 ## 4.4 Sharded File Content Access
 
 `getShards : PrivateFile -> Array<Namefilter>`
 
-[External content](#3211-externalized-content) namefilters are defined thus:
+To calculate the array of HAMT labels for [external content](#3211-externalized-content), add `key` and `sha3(key || encode(i))` for each block index `i` of external content to the `bareName` like so:
```
expede, line 320: 🎉 🧮

expede (approved):
Just a few questions and suggestions. The BFT one is the big open question for me, but it's relatively minor / shouldn't prevent progress in the meantime.

Lookin' good overall!

spec/private-wnfs.md
matheus23: I've tried to answer this question at the end of this section: 

> If any externalized content blocks exceed the specified `blockSize` or are missing in the private forest despite having a lower index than `blockCount` during file read operations, then these operations MUST produce an error.

The idea is to keep this invariant: Two clients should never have conflicting 'ideas' of what the data is. (Admittedly very hand-wavey)
The thinking is this: If we didn't have a count, how would we determine the end of a file?
Probably by reading blocks until we hit the first non-complete or missing block.
This works if we force everyone to start reading files from the beginning, but if someone wants to skip the first 10 blocks and read from then on, they may 'detect' a different end-of-file than someone reading from the beginning, if let's say the 9th block is missing.
I really want to prevent two clients thinking they've detected different file sizes. I'd rather have reads fail when they hit missing blocks. So essentially this makes the block count *the* source of truth.

BFT-ness is really tricky when we're doing things lazily. I guess you can only guarantee BFT-ness if you validate your data-structures deeply, right? I still need to wrap my head around that.

(By the way - you've asked this question the [last time you reviewed this](https://github.com/wnfs-wg/spec/pull/44#discussion_r1013436580) :sweat_smile: )

spec/private-wnfs.md
matheus23: Sounds better! I'm applying the `the section` removal, but I'll keep the period :stuck_out_tongue:

spec/private-wnfs.md
matheus23: I like it! I'll go for `consists of`, instead of `contains the following children` :)

matheus23: Ah right, and this is somewhat awkward, since this spec actually defines an abstract `HAMT` data structure and then applies it to `Namefilter`s as keys and `Array<CID>`s as values to and names the applied version the `PrivateForest`.

I'll adjust accordingly. In the future I think we'll probably be able to just 'refer to' the [IPLD HAMT spec](https://github.com/ipld/ipld/tree/master/specs/advanced-data-layouts/hamt).

spec/private-wnfs.md
matheus23: I'm not 100% sure whether we should require exceptions to be consistent across implementations, but I'm also not super against it.

I've added this paragraph specifically to address the question of what the source of truth: `blockSize` and `blockCount`are.

matheus23:
> The BFT one is the big open question for me, but it's relatively minor / shouldn't prevent progress in the meantime.

Yeah. I've responded to your comment on the line! Please take a look.

spec/private-wnfs.md
expede: >  how would we determine the end of a file?

Exponential search!

matheus23: Why not a count? That's faster. You just know the end of the file.

Exponential search has the same problems I described above.

Issue #43, opened 2022-10-17
Shared Private Files


matheus23:
Previous writing: https://whitepaper.fission.codes/file-system/shared

Needs some thinking about how to reconcile this with conflict resolution.

icidasset:
Do we have any thoughts yet on a spec for the format of the softlinks (aka. symbolic links) that point to the filesystem of the sharer?

Current format:
```js
{
  ipns: $dns_link_address,
  name: "Link name",
  privateName: $private_namefilter,
  key: $base64pad_encrypted_aes_key
}
```

matheus23:
Nope, that's still TBD (see #24).

It's probably going to be mostly like that, except it won't contain the "name", and maybe there's going to be a distinction between the content key and the revision key (whether the link is giving access to all future revisions or only a single revision).

matheus23:
Actually, shared private files in this version of the spec doesn't depend on softlinks at all. If one wants to group a bunch of things for sharing together, one can just create a directory and share that (and an application may do that on-the-fly if that should be hidden from the user).

Also, it's likely that softlinks won't be a mechanism for key management like it was the case before. I.e. we used to have keys in symlinks as you note above, but the next version will likely not have that.

Anyway. I other news, the spec was just merged, so we can close this :)

PR #51, opened 2022-11-25
Shared private data

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/shared/spec/shared-private-data.md)

Previously: https://whitepaper.fission.codes/file-system/shared

### DIFF
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -8,7 +8,7 @@ We need to store a list of all exchange keys, publicly readable, attached to a u
 
 We've decided to put this list under `exchange` at the root of the file system.
 
-## Alternatives
+## Alternatives Considered
 
 We've previously put these exchange keys at `/public/.well-known/exchange/*`. However, it may not be clear to users that giving another person or app write access to the root of the file system allows them to receive private files shared with this user.
 
@@ -16,4 +16,20 @@ Extracting these keys into its own section avoids this & encourages implementati
 
 # RSA Keys
 
-TODO (Essentially 'Because WebCrypto & non-extractability')
+## Need
+
+We need a way to deposit secrets for a known user non-interactively.
+
+## Choice
+
+We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)) mainly because it is possible to use non-extrcatable RSA keys in the WebCrypto API, thus keeping these keys away from potentially malicious extensions or app updates.
+
+## Alternatives Considered
+
+ECDH. The advantages ought to be:
+- Authentication of the sender
+- Smaller key sizes
+
+We don't need authentication for shared private files, as that's already covered by file system modification authorization systems outside WNFS (e.g. through UCANs).
+
+It also complicates looking up deposited messages for receivers, since they need to go through all combinations of sender & receiver public keys.
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -22,7 +22,7 @@ We need a way to deposit secrets for a known user non-interactively.
 
 ## Choice
 
-We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)) mainly because it is possible to use non-extrcatable RSA keys in the WebCrypto API, thus keeping these keys away from potentially malicious extensions or app updates.
+We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)) mainly because it is possible to use non-extractable RSA keys in the WebCrypto API, thus keeping these keys away from potentially malicious extensions or app updates.
 
 ## Alternatives Considered
 
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -1,0 +1,19 @@
+# Exchange Key Location
+
+## Need
+
+We need to store a list of all exchange keys, publicly readable, attached to a user's file system.
+
+## Choice
+
+We've decided to put this list under `exchange` at the root of the file system.
+
+## Alternatives
+
+We've previously put these exchange keys at `/public/.well-known/exchange/*`. However, it may not be clear to users that giving another person or app write access to the root of the file system allows them to receive private files shared with this user.
+
+Extracting these keys into its own section avoids this & encourages implementations to use a special UCAN capability for managing write access to this list of keys.
+
+# RSA Keys
+
+TODO (Essentially 'Because WebCrypto & non-extractability')
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -41,3 +41,5 @@ It also complicates looking up deposited messages for receivers, since they need
 In contrast to RSAES-OAEP, this is a padding-less encryption scheme, which makes implementations easier and is supposedly faster.
 
 However, as browsers are one of our target environments and only RSAES-OAEP is currently widely supported in WebCrypto, we can't use RSA KEM-DEM.
+
+Should browsers widely support e.g. X25519-based ECDH, we're happy to start working on a new protocol version based on such keys.
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -40,4 +40,4 @@ It also complicates looking up deposited messages for receivers, since they need
 
 In contrast to RSAES-OAEP, this is a padding-less encryption scheme, which makes implementations easier and is supposedly faster.
 
-However, as browsers are one of our target environments and only RSA-OAEP is currently widely supported in WebCrypto, we can't use RSA KEM-DEM.
+However, as browsers are one of our target environments and only RSAES-OAEP is currently widely supported in WebCrypto, we can't use RSA KEM-DEM.
--- a/rationale/shared-private-data.md
+++ b/rationale/shared-private-data.md
@@ -26,10 +26,18 @@ We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker
 
 ## Alternatives Considered
 
-ECDH. The advantages ought to be:
+### ECDH
+
+The advantages ought to be:
 - Authentication of the sender
 - Smaller key sizes
 
 We don't need authentication for shared private files, as that's already covered by file system modification authorization systems outside WNFS (e.g. through UCANs).
 
 It also complicates looking up deposited messages for receivers, since they need to go through all combinations of sender & receiver public keys.
+
+### [RSA KEM-DEM](https://www.rfc-editor.org/rfc/rfc5990.html#appendix-A)
+
+In contrast to RSAES-OAEP, this is a padding-less encryption scheme, which makes implementations easier and is supposedly faster.
+
+However, as browsers are one of our target environments and only RSA-OAEP is currently widely supported in WebCrypto, we can't use RSA KEM-DEM.
--- a/spec/shared-private-data.md
+++ b/spec/shared-private-data.md
@@ -1,0 +1,104 @@
+# WNFS Shared Private Data Extension Specification
+
+# 0 Abstract
+
+There are several cases where we need to exchange private data asynchronously when the other party may be offline.
+The shared private data extension allows a file system's owner to deposit messages encrypted with a recipient's public key on their private forest.
+A recipient will then periodically - or after being prompted - scan the sender's private forest for payloads they can decrypt.
+These encrypted payloads contain secrets giving read access and/or [UCANs](https://ucan.xyz) giving write access to parts of the owner's public or private file system.
+
+# 1 Exchange Keys
+
+To share information with a user that's offline we make use of asymmetric encryption. All WNFS users widely distribute a list of public 2048-bit RSA public keys - their non-exportable "exchange keys" - as `did:key:` DIDs at a well-known location: A list under `exchange` at the root of WNFS next to `public` and `private`. See [the rationale](/rationale/shared-private-data.md#exchange-key-location) for more information on this.
+
+These RSA keys are used to encrypt a symmetric key for a recipient; the "share key". This symmetric key is then used to decrypt a pointer a private node.
+
+Allowing multiple exchange keys per WNFS enables multiple recipient devices to receive a share without having to transfer exchange keys between devices.
+
+# 2 Data Layout
+
+Shared private data payloads are stored in the [Private Forest](/spec/private-wnfs.md#21-ciphertext-blocks).
+
+## 2.1 Share Label
+
+Shares are labeled in the private forest by a namefilter consisting of:
+- The sender's root DID. In most cases this will be the file system owner's root DID.
+- The recipient's exchange DID
+- A counter, which is incremented each time the sender creates a share.
+
+The counter MUST be a 0-based natural number.
+
+The namefilter for a share is computed like this:
+
+```ts
+function computeShareLabel(senderRootDID: string, recipientExchangeDID: string, counter: number) {
+  return new Namefilter()
+    .add(senderRootDID)
+    .add(recipientExchangeDID)
+    .add(counter.toString())
+    .saturate()
+}
+```
+
+## 2.2. Share Payload
+
+The share payload MUST be a non-empty list of CIDs to [RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1) encrypted ciphertexts.
+
+When decrypted, this reveals a payload of the form:
+
+`<unsigned varint prefix><32 byte private forest label><32 byte AES key>`
+
+As exchange keys are 2048 bit and can thus encrypt only up to 190 bits of data, the unsigned varint prefix MUST be smaller than $190 - 32 \cdot 2 = 126$ bytes.
+
+In the current specification, valid prefixes always encode into 8 bytes. This means *decrypted* share payloads will always be 72 bytes long.
+
+Share payload *ciphertexts* will always be 256 bytes long, due to the way RSAES-OAEP works.
+
+### 2.2.1 Unsigned Varint Prefix
+
+WNFS (currently) uses the [multicodec private use range](https://github.com/multiformats/multicodec/#private-use-area) for prefixing and identifying share pointers.
+
+| Prefix | Meaning | Decoded Unsigned Varint | Prefix Computation |
+|-------:|---------|-------------------------|:-------------------|
+| `0xceaec101` | Temporal Share Pointer | `0x30574e` | `varint_enc(0x300000 \| ascii_enc("WN"))` |
+| `0xcfaec101` | Snapshot Share Pointer | `0x30574d` | `varint_enc(0x300000 \| ascii_enc("WN") + 1)` |
+
+### 2.2.2. Temporal Share Pointer
+
+After the unsigned varint prefix `0xceaec101`, the temporal share pointer consists of
+- a 32 byte private forest label used to look up the private node to decrypt in the private forest
+- a 32 byte [revision key](/spec/private-wnfs.md#3161-revision-key) to decrypt the private node and its temporal header.
+
+### 2.2.3 Snapshot Share Pointer
+
+The snapshot share pointer works exactly like the temporal share pointer, except it is prefixed with `0xcfaec101` and instead of encoding a revision key, it encodes a [content key](/spec/private-wnfs.md#3162-content-key).
+
+This content key only gives access to a single revision, in case a user wanted to only share a single revision with someone else.
+
+# 3 Share Lookup and Discovery
+
+It's not assumed that recipients are notified over some channel about whether they recieved any new shares.
+
+Thus it is possible to scan other user's file systems for files that were shared with a given exchange key. To do this, generate share labels as described in [section Share Label](#21-share-label) starting from 0 or the last counter used for lookup until the first missing label in the private forest is hit.
+
+# 4 Concurrent Share Creation
+
+If multiple shares are created concurrently on separate devices, then these can simply be merged according to the [private forest merge algorithm](/spec/private-wnfs.md#45-merge).
+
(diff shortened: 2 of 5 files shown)


matheus23:
Closes #43

rationale/shared-private-data.md
```diff
@@ -0,0 +1,35 @@
...
+
+We need to store a list of all exchange keys, publicly readable, attached to a user's file system.
+
+## Choice
+
+We've decided to put this list under `exchange` at the root of the file system.
+
+## Alternatives Considered
+
+We've previously put these exchange keys at `/public/.well-known/exchange/*`. However, it may not be clear to users that giving another person or app write access to the root of the file system allows them to receive private files shared with this user.
```
QuinnWilton, line 13: This is a __very__ good point, with possible implications for other use-cases for `.well-known`. Flagging it for @blaine in case it raises any concerns around any of their ideas around naming and identity.

spec/shared-private-data.md
```diff
@@ -0,0 +1,104 @@
...
+There are several cases where we need to exchange private data asynchronously when the other party may be offline.
+The shared private data extension allows a file system's owner to deposit asymmetrically encrypted messages so they can be picked up by the recipient when they come online.
+A recipient will then periodically - or after being prompted - scan the sender's private forest for payloads they can decrypt.
+These encrypted payloads contain secrets giving read access and/or [UCANs](https://ucan.xyz) giving write access to parts of the owner's public or private file system.
+
+# 1 Exchange Keys
+
+To share information with a user that's offline we make use of asymmetric encryption. All WNFS users widely distribute a list of 2048-bit RSA public keys - their non-exportable "exchange keys" - as `did:key:` DIDs at a well-known location: A list under `exchange` at the root of WNFS next to `public` and `private`. See [the rationale](/rationale/shared-private-data.md#exchange-key-location) for more information.
+
+These RSA keys are used to encrypt a share payload for a recipient. This share payload contains a pointer to a private node and the symmetric key to decrypt it.
```
QuinnWilton, line 14: It wasn't initially clear to me that "share payload" formed a noun phrase, rather than a typo. Perhaps it'd help if you linked it to the section: https://github.com/wnfs-wg/spec/blob/matheus23/shared/spec/shared-private-data.md#22-share-payload

spec/shared-private-data.md
```diff
@@ -0,0 +1,104 @@
...
+
+Allowing multiple exchange keys per WNFS enables multiple recipient devices to receive a share without having to transfer exchange keys between devices.
+
+# 2 Data Layout
+
+Shared private data payloads are stored in the [Private Forest](/spec/private-wnfs.md#21-ciphertext-blocks).
+
+## 2.1 Share Label
+
+Shares are labeled in the private forest by a namefilter consisting of:
```
QuinnWilton, line 24: I'd also consider linking to the [namefilter spec](https://github.com/wnfs-wg/spec/blob/matheus23/shared/spec/namefilter.md) here :)

matheus23:
Also I would like to know what people's thoughts are on using `did:key:` with our WNFS exchange keys. Exchange keys are *only* used for en/decryption, never for signing. However, IIUC, DIDs are meant to only be used for signatures. It seems somewhat like a misuse to have DIDs in here. Should we just drop the `did:key:` prefix and make these multibase-formatted multiformat public keys?

matheus23:
Some TODOs from reviews on other channels:
- [ ] Add a protocol version to exchange keys
- [ ] Consider grouping exchange keys of different protocol versions by device (senders should not have to encrypt payloads for 20 different protocol versions that some devices supports. Using the most recent version it knows should be sufficient)
- [ ] *Perhaps* consider specifying ECDH-based exchange keys & share payloads.
- [ ] Be more precise & fix RSA params. E.g. 65537 exponent!

icidasset:
> Also I would like to know what people's thoughts are on using did:key: with our WNFS exchange keys

I only see it making sense for the usage in `public/.well-known/`. We could possibly have other DID methods used in there as well (eg. `did:dns:`) to make sharing easier, depending on the use case.

rationale/shared-private-data.md
```diff
@@ -0,0 +1,35 @@
...
+
+# RSA Keys
+
+## Need
+
+We need a way to deposit secrets for a known user non-interactively.
+
+## Choice
+
+We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)) mainly because it is possible to use non-extractable RSA keys in the WebCrypto API, thus keeping these keys away from potentially malicious extensions or app updates.
```
expede, line 25: And because we've historically been deeply paranoid about the controversy around the NIST curves 😅

dignifiedquire: Any chance we can avoid building new systems that require RSA? 😭

matheus23: WebCrypto gives us RSA-OAEP and ECDH with P255.
Once WebCrypto gives us ECDH with ed25519, we'll switch to that.

expede: @dignifiedquire unfortunately RSA is ubiquitous today. As @matheus23 says, there's nothing special about RSA here other than it works in browsers and is not the NIST curves. Every few months I go "how bad could the NIST curves really be?" do some digging, sigh deeply, and bite the bullet with RSA (for now).

dignifiedquire: > and bite the bullet with RSA

The challenge is that RSA ends up being insecure in a lot of cases, even without NIST doing any work 😓 due to the challenges in implementing it securely

matheus23:
Made some changes. The summary is:


- The share protocol is now versioned, instead of using crypto agility via multicodecs.
- We no longer use DIDs as encodings for exchange keys. The initial protocol will just encode the RSA public modulus as 256 lower-endian bytes.
- The exchange partition is now simply its own public WNFS data structure, but intentionally separated from public files. This allows it to inherit the merge algorithm.

rationale/shared-private-data.md
zeeshanlakhani: Generally, a `Recommendations` section is generally good for specs in the large.

rationale/shared-private-data.md
```diff
@@ -0,0 +1,43 @@
+# Exchange Key Location
```
zeeshanlakhani, line 1: generally, a big fan of this rationale doc, as it brings the RFC feel closer to the spec. We should do this more often across our specs.

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
+# WNFS Shared Private Data Extension Specification
+
+# 0 Abstract
+
+There are several cases where we need to exchange private data asynchronously when the other party may be offline.
```
zeeshanlakhani, line 5: Can we list some example cases? If not here, in the rationale?

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+The shared private data extension allows a file system's owner to deposit asymmetrically encrypted messages so they can be picked up by the recipient when they come online.
+A recipient will then periodically - or after being prompted - scan the sender's private forest for payloads they can decrypt.
```
zeeshanlakhani, line 7: Would like recommendation(s) on how this can be done periodically or after being prompted?

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+As exchange keys are 2048 bit and can thus encrypt only up to 190 bits of data, the [unsigned varint](https://github.com/multiformats/unsigned-varint) prefix MUST be smaller than $190 - 32 \cdot 2 = 126$ bytes.
+
+In the current specification, valid prefixes always encode into 8 bytes. This means *decrypted* share payloads will always be 72 bytes long.
+
+Share payload *ciphertexts* will always be 256 bytes long, due to the way RSAES-OAEP works.
+
+### 3.2.1 Unsigned Varint Prefix
+
+WNFS (currently) uses the [multicodec private use range](https://github.com/multiformats/multicodec/#private-use-area) for prefixing and identifying share pointers.
```
zeeshanlakhani, line 99: I see this in the repo, but is this range canonical in a spec somewhere?

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+The snapshot share pointer works exactly like the temporal share pointer, except it is prefixed with `0xcfaec101` and instead of encoding a revision key, it encodes a [content key](/spec/private-wnfs.md#3162-content-key).
+
+This content key only gives access to a single revision, in case a user wanted to only share a single revision with someone else.
+
+# 4 Share Lookup and Discovery
+
+It's not assumed that recipients are notified over some secondary channel when they receive new shares.
+
+Thus it is possible to scan other user's file systems for files that were shared with a given exchange key. To do this, generate share labels as described in [section Share Label](#21-share-label) starting from 0 or the last counter used for lookup until the first missing label in the private forest is hit.
```
zeeshanlakhani, line 124: share label anchor?

spec/shared-private-data.md
matheus23: Hm. Examples would include stuff like "device owner is asleep", "device is turned off" or "device is not connected to the internet".

I think perhaps instead I can motivate this via store-and-forward networks?

spec/shared-private-data.md
matheus23: I'm pretty sure that repository is the spec, the source of truth. It kind of sucks.
There used to be some effort in getting an RFC into IETF: https://github.com/multiformats/specs
But that seems to have faded out 2 years ago. Also, the [multiformats website](https://multiformats.io/) just links the repository.

rationale/shared-private-data.md
expede: Yeah a section to this effect (IMO in the same section) would be really helpful. Agreed that `RECOMMEND` or `SHOULD` is the right level.

rationale/shared-private-data.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Choice
+
+We've decided to use asymmetric RSA encryption ([RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)) mainly because it is possible to use non-extractable RSA keys in the WebCrypto API, thus keeping these keys away from potentially malicious extensions or app updates.
+
+## Alternatives Considered
+
+### ECDH
+
+The advantages ought to be:
```
expede, line 31: Is this document meant to conform to RFC2119?

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+# 0 Abstract
+
+There are several cases where users want to exchange private data asynchronously when the other party may be offline. In such cases we can make use of [store-and-forward networks](https://en.wikipedia.org/wiki/Store_and_forward) to deposit data for later retrieval.
+The shared private data extension allows a file system's owner to deposit asymmetrically encrypted messages so they can be picked up by recipients when they come online.
+Recipients can periodically - or after being prompted - scan the sender's private forest for payloads they can decrypt, for example by looking for updates in file systems from a contact list, or by being sent a message pointing to the share payload over an out-of-bounds secure channel.
+These encrypted payloads contain secrets giving read access and/or [UCANs](https://ucan.xyz) giving write access to parts of the owner's public or private file system.
+
+# 1 Protocol Versions
+
+The shared private data extension is versioned and designed to support multiple versions living on the same file system concurrently, so devices can gradually upgrade to the new version and support multiple versions at the same time.
```
expede, line 12: Can you define "extension" in the context of this spec?

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+Version 1 of the shared private data protocol uses RSA public keys for asymmetric encryption. These RSA public keys MUST have an exponent of 65537 and MUST be 2048-bit.
+The method for encryption MUST be RSAES-OAEP.
+
+# 2 Exchange Keys Partition
+
+To share information with a user who is offline we make use of asymmetric encryption. All WNFS users widely distribute a list of public keys - their non-exportable "exchange public keys" at a well-known location: A map under `exchange` at the root of WNFS next to `public` and `private`. See [the rationale](/rationale/shared-private-data.md#exchange-key-location) for more information.
+
+These keys are used to encrypt a [share payload](#32-share-payload) for a recipient. This share payload contains a pointer to a private node and the symmetric key to decrypt it.
+
+Allowing multiple exchange keys per WNFS enables multiple recipient devices to receive a share without having to transfer exchange keys between devices.
```
expede, line 27: > without having to transfer exchange keys between devices.

It may be worth adding a sentence or two about why this is important

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+## 2.1 Layout
+
+The exchange key partition is a link to a [public directory](/spec/public-wnfs.md). This directory contains a set of further directories with exchange keys within.
```
expede, line 31: Not sure if it's just me, but this took me a couple times to realize that it didn't mean that you're linking to the public partition

```suggestion
The exchange key partition is structured as a [public directory](/spec/public-wnfs.md) under its own top-level link. This directory contains further directories containing exchange keys.
```

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+Exchange keys are grouped by device, such that a sender only needs to choose *one* of the available keys for each device. For example, senders that support protocol version 1 and 2 look through all recipient's devices and select either a single version 2 exchange key or, if no such key is present, a version 1 key from each device.
```
expede, line 33: Why have it broken down by device rather than a list of public keys per version? So that they can be revoked by context later as needed?

I would also perhaps change the term from "device" to "agent" or "content". e.g. laptop Chrome vs laptop Firefox vs native.

expede: (I should have highlighted lines 33-38 for the above)

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+## 3.1 Share Label
+
+Shares are labeled in the private forest by a [namefilter](/spec/namefilter.md) consisting of:
+- The sender's root DID. In most cases this will be the file system owner's root DID.
+- The recipient's exchange key
+- A counter, which is incremented each time the sender creates a share.
+
+The counter MUST be a 0-based natural number.
+
+The namefilter for a share is computed like this:
```
expede, line 65: If following RFC 2119:

```suggestion
The namefilter for a share MUST be computed as follows:
```

spec/shared-private-data.md
expede: From later, this is broken down as:

```
/exchange
├── aNXmyZ-kRI1-fsLcOol8wi0wfmxxY5_15bxdt2T4bs8
│   ├── v1.exchangekey
│   ├── v2.exchangekey
├── BEubij6fgCV51wyhYtITYCYhoGvk5Y5Slr59Kl4FOkQ
│   └── v1.exchangekey
└── zulwKufD7v71rJ7792cl2pi1LHmePS7d7a0vhuvaKNo
    └── v2.exchangekey
```

...why not...

```
/exchange
├── v1
│   ├── publicKey1
│   └── publicKey2
└── v2
    ├── publicKey3
    └── publicKey4

```

...or...

```
/exchange
├── v1.json
└── v2.json
```

...where the values of the `.json` is dag-json broken out into separate blocks (and thus can easily be checked to be append-only).

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+## 3.2. Share Payload
+
+The share payload MUST be a non-empty list of CIDs to [RSAES-OAEP](https://datatracker.ietf.org/doc/html/rfc3447#section-7.1)-encrypted ciphertexts.
+
+When decrypted, this reveals a payload of the form:
+
+`<unsigned varint prefix><32 byte private forest label><32 byte AES key>`
+
+This is a [multicodec](https://github.com/multiformats/multicodec/).
```
expede, line 89:
```suggestion
The decrypted payload MUST be represented as a [multicodec](https://github.com/multiformats/multicodec/) in the form:

`<unsigned varint prefix><32 byte private forest label><32 byte AES key>`
```

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+As exchange keys are 2048 bit and can thus encrypt only up to 190 bits of data, the [unsigned varint](https://github.com/multiformats/unsigned-varint) prefix MUST be smaller than $190 - 32 \cdot 2 = 126$ bytes.
+
+In the current specification, valid prefixes always encode into 8 bytes. This means *decrypted* share payloads will always be 72 bytes long.
+
+Share payload *ciphertexts* will always be 256 bytes long, due to the way RSAES-OAEP works.
+
+### 3.2.1 Unsigned Varint Prefix
+
+WNFS (currently) uses the [multicodec private use range](https://github.com/multiformats/multicodec/#private-use-area) for prefixing and identifying share pointers.
```
expede, line 99: Awesome 👍 

Not for _right now,_ but these kinds of decryption pointers have come up in a couple contexts. if we haven't already, it may be worth us writing a standard for submission to multiformats.

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+### 3.2.2. Temporal Share Pointer
+
+After the unsigned varint prefix `0xceaec101`, the temporal share pointer consists of
+- a 32 byte private forest label used to look up the private node to decrypt in the private forest
+- a 32 byte [revision key](/spec/private-wnfs.md#3161-revision-key) to decrypt the private node and its temporal header.
+
+### 3.2.3 Snapshot Share Pointer
+
+The snapshot share pointer works exactly like the temporal share pointer, except it is prefixed with `0xcfaec101` and instead of encoding a revision key, it encodes a [content key](/spec/private-wnfs.md#3162-content-key).
```
expede, line 116: Purely out of curiosity: is there a pun or some other meaning behind `0xcfaec101`?

expede: Nevermind, it's above... this is what I get for bouncing between tasks 😅

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+| `0xceaec101` | Temporal Share Pointer | `0x30574e` | `varint_enc(0x300000 \| ascii_enc("WN"))` |
+| `0xcfaec101` | Snapshot Share Pointer | `0x30574d` | `varint_enc(0x300000 \| ascii_enc("WN") + 1)` |
+
+The `varint_enc` function encodes an [unsigned varint](https://github.com/multiformats/unsigned-varint).
+
+### 3.2.2. Temporal Share Pointer
+
+After the unsigned varint prefix `0xceaec101`, the temporal share pointer consists of
+- a 32 byte private forest label used to look up the private node to decrypt in the private forest
+- a 32 byte [revision key](/spec/private-wnfs.md#3161-revision-key) to decrypt the private node and its temporal header.
```
expede, line 112: 👍

spec/shared-private-data.md
```diff
@@ -0,0 +1,190 @@
...
+
+### 3.2.3 Snapshot Share Pointer
+
+The snapshot share pointer works exactly like the temporal share pointer, except it is prefixed with `0xcfaec101` and instead of encoding a revision key, it encodes a [content key](/spec/private-wnfs.md#3162-content-key).
+
+This content key only gives access to a single revision, in case a user wanted to only share a single revision with someone else.
```
expede, line 118: Interesting. Do we have a similar convention for all decryption pointers?

spec/shared-private-data.md
matheus23: I'm thinking of it like this: The rest of the spec doesn't have a dependency on shared private data existing. It's technically possible to ship an implementation of WNFS without shared private data and it'd work and be useful.
Not super attached to this idea, it was basically a "why not?".
Should I add a paragraph explaining this? Or should I rather just make shared private data part of the WNFS spec?

spec/shared-private-data.md
matheus23: Yeah! Thanks, that's better.

spec/shared-private-data.md
matheus23: > Why have it broken down by device rather than a list of public keys per version?

Because when creating share payloads, you want to pick *one key for each device*.
If you structure it differently, you lose the mapping of which groups of keys belong to one device. Then you're forced to encrypt for *every* exchange key, and since some devices advertise both v1 and v2 keys, you'll do some redundant work.

I tried to explain that reasoning with the paragraph in line 33.

spec/shared-private-data.md
matheus23: Yes. I want to play with them a bit, see where else they end up useful and plan to do exactly that when it feels useful.

spec/shared-private-data.md
matheus23: Assuming you're referring to the unsigned varint prefixing, that's only for shared private files.
Signaling generally works differently for the rest of WNFS: We're using `type` keys in CBOR objects.
I initially wanted to do the same in this spec. However, I have to save bytes as I can only encrypt up to 190 bytes using RSA-OAEP. Using the unsigned varint and a packed byte format saves some bytes.

matheus23: Perhaps that's premature optimization. CBOR is fairly concise, too. A CBOR encoding would fit within <110 bytes.

expede: From our call: this should be consistent with the private partition, whether that's by file name or multicodec prefix.

spec/shared-private-data.md
expede: Ah, I see. Arguably the same can be said for private and public data not having dependencies on each other. If we want to go down the path of something being totally optional, it probably should live in its own spec with WNFS as a dependency. I would be surprised if someone _didn't_ want to implement this as a feature in an application, though at the spec level there's definitely an argument to be made.

expede (approved):
Just the one comment on consistency, but otherwise LGTM! 🚀✨

matheus23:
Went for cbor-based signaling for consistency with the rest of WNFS, instead of adding our own multicodecs.

PR #52, opened 2022-12-01
Symlinks

### LINKED ISSUE #24: Symlink Spec

matheus23:
Closes #24 

[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/symlinks/spec/symlinks.md)

TODO:
- [ ] How would you import relative unix symlinks into WNFS? In public WNFS we already have a path, so it's fairly straight-forward to support, in private WNFS though? Would it not refer to an `inumber`?

### DIFF
--- a/spec/symlinks.md
+++ b/spec/symlinks.md
@@ -38,7 +38,7 @@ type PublicSymlink = {
 }
 ```
 
-Public data is symlinked via a list of utf8 path segment strings.
+Public data is symlinked via a list of [UTF-8](https://www.rfc-editor.org/rfc/rfc3629) path segment strings.
 
 If the symlink is not linking relative to the same file system, it can provide a `root` parameter: An [`ipns://` URI](https://github.com/ipfs/in-web-browsers/blob/master/ADDRESSING.md) to the public root that the path is relative to.
 
--- a/spec/symlinks.md
+++ b/spec/symlinks.md
@@ -1,0 +1,40 @@
+# Symlink Specification
+
+# 0 Abstract
+
+Symlinks are a tool for users to organize their file system:
+- They allow them to link to the same directory from different places in their file system
+- They enable linking out to other user's file systems from your own file system
+- They enable linking between public and private parts of file systems
+
+This shouldn't be confused with a key management mechanism: Knowing a symlink shouldn't give someone more read or write access. This prevents e.g. accidentally giving someone access to another file system's files if a directory includes a symlink.
+
+While hard links (`CID`s or private references) conceptually are supposed to work like 'copying' the linked data, a symlink is meant to work differently:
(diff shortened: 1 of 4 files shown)


spec/symlinks.md
```diff
@@ -0,0 +1,118 @@
...
+}
+```
+
+## 1.2 Private Symlink
+
+```ts
+type PrivateSymlink = {
+  type: "wnfs/sym/priv"
+  root?: string
+  inumber: Inumber
```
zeeshanlakhani, line 86: dumb q alert: guessing this Inumber is always unique (after looking at the current code too)? Do we have anything like an Inode for filesystems in wnfs? https://en.wikipedia.org/wiki/Inode.

spec/symlinks.md
```diff
@@ -0,0 +1,118 @@
...
+Symlink is short for "Symbolic Link".
+
+# 0 Abstract
+
+Symlinks are a tool for users to organize their file system:
+- They allow linking to the same directory from different places in their file system
+- They enable linking out to other user's file systems from your own file system
+- They enable linking between public and private parts of file systems
+
+While hard links (`CID`s or private references) conceptually are supposed to work like 'copying' the linked data, a symlink is meant to work differently:
```
zeeshanlakhani, line 12: code-wise, do you see any aliasing needed for hardlinks now (vs symlinks)?

spec/symlinks.md
matheus23: Yes, this number *should* be unique. You could think of files and directories at inodes. The term `inumber` also appears in the linked articles. It would then be the number that refers to a certain inode.
However, in our case the inumber is persistent across revisions. This is where the analogy kind-of breaks down. I'm not super sure whether inumbers change between revisions in traditional file systems, or whether everything basically happens via in-place mutation.

spec/symlinks.md
matheus23: No. I don't think so.

I'm still thinking idly about the symlinking stuff in WNFS.
Only in the public file system hard-links correspond directly to CIDs.
In the private file system, that's not quite the case.

If our "own" version of hard-links doesn't become a thing, this will probably be a comment on the `cp` function of the public file system, as it doesn't actually need to copy anything, it's similar to a hard link.

matheus23:
Closing this for now, as I don't intend to work on a spec in the near future.
The problems is mainly that there's a tug-of-war happening with symlinks where there's three different use-cases that I fear all warrant slightly different solutions:
- Deduplication (on the decrypted level)
- Key management (e.g. collecting access to all of the directories and files that were 'shared with you' from other users in a central place)
- POSIX file system round-tripping. I.e. injesting POSIX symlinks as WNFS and being able to output them again.

PR #57, opened 2023-01-10
Fix maximum block size calculation & some typos

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/cleanup/spec/private-wnfs.md)

While going through and fixing some typos, I noticed that the maximum plaintext bytes per ciphertext block calculation was off.
The overhead for AES-GCM encryption is 28 bytes: 16 bytes for the authentication tag and 12 bytes for the initialization vector.

Also, for some reason the calculation was just... wrong. $2^{18} - 12 \neq 262,322$!
Instead, it should be $2^{18} - 28 = 262,116$.

### DIFF
--- a/spec/private-wnfs.md
+++ b/spec/private-wnfs.md
@@ -176,7 +176,22 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-sharded-file-content-access) for more detail.
 
-The block size MUST be at least 1 and at maximum $2^{18} - 28 = 262,116$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes and 16 authentication tag bytes need to be added to each ciphertext. It is RECOMMENDED to use the maximum block size.
+The block size MUST be at least 1 and at maximum $2^{18} - 28 = 262,116$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes and 16 authentication tag bytes need to be added to each ciphertext. It is RECOMMENDED to use the maximum block size. An externalized content block is laid out like this:
+
+```
+ 0                   1
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| Initialization Vector |         |
++-+-+-+-+-+-+-+-+-+-+-+-+         |
+|                                 :
+:         Encrypted Block         :
+:        (blockSize bytes)        |
+|                                 |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|        Authentication Tag       |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
 
 The block count MUST reference the number of blocks the externalized content was split into.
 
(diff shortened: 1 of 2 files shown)


spec/private-wnfs.md
```diff
@@ -176,15 +176,15 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 Since external content blocks are separate from the header, they MUST have a unique namefilter derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-shared-file-content-access) for more detail.
```
zeeshanlakhani, line 177: `#44-shared-file-content-access` doesn't seem to link right? Should it be `#4.4-shared-file-content-access`?

matheus23: Thank you for clicking through all of these! (Is what I assume you're doing)

matheus23: Okay. Clicked through it all. So many broken anchor links.
I am *so much* not looking for doing this again when I've added a bunch of sections :upside_down_face:

spec/private-wnfs.md
```diff
@@ -176,15 +176,15 @@ Private file content has two variants: inlined or externalized. Externalized con
...
 
-The block size MUST be at least 1 and at maximum $2^{18} - 12 = 262,322$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes need to be prepended to each ciphertext. It is RECOMMENDED to use the maximum block size.
+The block size MUST be at least 1 and at maximum $2^{18} - 28 = 262,116$ bytes, as the maximum block size for IPLD is usually $2^{18}$, but 12 initialization vector bytes and 16 authentication tag bytes need to be added to each ciphertext. It is RECOMMENDED to use the maximum block size.
```
zeeshanlakhani, line 179: I also looked at the code, but is there a reference/spec we can link to for the 12 initialization vector bytes and 16 authentication tag bytes that need to be added? It may be good to link the max block size for IPLD too (in its own spec). 

When I did packet integrations, we ascii'ed out the bytes for headers/segments/layers, but it may be nice to include that somewhere?

matheus23: > When I did packet integrations, we ascii'ed out the bytes for headers/segments/layers, but it may be nice to include that somewhere?

That sounds like a great idea.

Agree about linking to specs about this, too, I'm unfortunately not aware of some good resource to point to for AES-GCM. Most stuff is PDFs of NIST documents, there's no RFC for AES-GCM AFAIK.
There's also no good resources for the IPFS block size limit. It's the limit that kubo gives itself when generating UnixFS blocks by default. However, kubo only starts rejecting blocks at soon as they hit 2MiB at the bitswap level IIUC. There may be other limits involved in the system.
So in short: It's complicated, I don't think there's a definitive answer to "what is the IPFS block size" and we're just using the most conservative number, matching UnixFS implementations.

zeeshanlakhani: On the PDFs of NIST, just a screenshot or again, some minor diagramming is good enough.

This bit about kubo limits seems like a great footnote on the spec :).

matheus23: Added a diagram.

Also, turns out that the NIST spec actually doesn't specify where to put the authentication tag. It's entirely up to the application developer, but most libraries just append the auth tag to the ciphertext automatically, e.g. that's what WebCrypto and the RustCrypto libs do.

zeeshanlakhani:
a couple of minor things:

* https://github.com/wnfs-wg/spec/blob/matheus23/cleanup/spec/private-wnfs.md#45-merge references non-rendered LaTeX/mathjax, i.e. ($\land$). 
* The `See [rationale/hamt.md](https://github.com/rationale/hamt.md) for more information.` footnote link ([1]) doesn't seem to anchor right.

matheus23:
> The See footnote link ([1]) doesn't seem to anchor right.

Fascinating. Seems like footnotes don't get the usual link treatment from github. I'll turn the footnote back into a sentence inside the spec. Otherwise it's impossible to link relative to the current branch.

Thanks for these :)

matheus23:
This is only
- Fixing typos
- Fixing a calculation and
- Adding a diagram

I'll just go ahead and merge this :v:

Issue #56, opened 2023-01-09
Private file system: Hierarchical pointer disambiguation


matheus23:
Needed to fully support #34

walkah:
Being handled as part of #47

PR #58, opened 2023-01-11
Nonnested encryption, disambiguated child pointers & deduplicate using deterministic encryption

matheus23:
[:scroll: Preview `private-wnfs.md`](https://github.com/wnfs-wg/spec/blob/matheus23/nonnested-encryption/spec/private-wnfs.md)
[:scroll: Preview `notation.md`](https://github.com/wnfs-wg/spec/blob/matheus23/nonnested-encryption/spec/notation.md)

We're reworking the data format that was introduced with the distinction between temporal and snapshot access (#6):

Due to different access levels, we needed to encrypt some parts of a private node with a different key than other parts. Previously we decided to nest the encryption, so it all stays "within the same block". This turns out to be a bad idea: It doesn't save us from having to refer to the specific block CID we intend to link to anyway (#56). If we need to do this either way, we can just split the block into multiple and use their CID as an additional reference disambiguation mechanism.

So this also addresses #56.

If we split out the private node header into its own block, we can also make use of deterministic encryption schemes like AES-GCM-SIV for private header nodes. This causes them not to be stored redundantly. We don't need IND-CPA for private node headers anyway.


spec/notation.md
```diff
@@ -3,25 +3,33 @@
...
+This represents a ciphertext that's marked with the data type `Data` that it needs to decrypt to. For example `AesGcm<Ratchet>` represents a byte string that would decrypt to a value of type `Ratchet` using [AES-GCM](https://csrc.nist.gov/publications/detail/sp/800-38d/final) with 256-bit keys.
+The ciphertexts produced by `AesGcm<>` have a 12 byte initialization vector prepended and the 16 byte authentication tag appended.
+
+## `AesKwp<Data>`
+
+This represents a ciphertext that decrypts to given data type `Data`.
+The encryption/decryption algorithm used is [AES-KWP](https://www.rfc-editor.org/rfc/rfc5649) with 256-bit keys.
+AES-KWP can be thought of as keyed permutation function where observing the permutation doesn't reveal information about the key used, together with an 8-byte authentication tag. It doesn't provide [IND-CCA2](https://en.wikipedia.org/wiki/Ciphertext_indistinguishability), so must only be only be used in cases where detecting two same-key and same-message ciphertexts is not a security concern.
+In this specification we use AES-KWP for encrypting random AES keys and private node headers containing the `inumber`, namefilter and skip ratchet.
```
expede, line 33: I'll say it every time: this file is super useful 🚀

expede: In a few places I think that you're using the `||` convention for concatenation. In code blocks my brains reads that as `OR` (though in $\LaTeX$ having $a \|| b$ is clearer thanks to the longer lines). Assuming that you mean `||` as concat, it may be worth noting that here 🤔

spec/private-wnfs.md
```diff
@@ -33,7 +33,9 @@ The encrypted layer is intended to hide as much information as possible, while s
 
 At the encrypted data layer, the private forest is a collection of ciphertext blocks. These blocks SHOULD be smaller than 256 kilobytes in order to comply with the default IPFS block size. Keeping block size small is also useful for reducing metadata leakage - it's less obvious what the file size distribution in the private file system is like if these files are split into blocks.
 
-Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multimap](https://en.wikipedia.org/wiki/Multimap). The HAMT MUST have a node-degree of 16, and MUST used saturated  saturated [namefilter](/spec/namefilter.md)s as keys. See [`rationale/hamt.md`](/rationale/hamt.md) for more information on parameter choice.
+We refer to the keys in the private forest as 'labels' to disambiguate them from cryptographic keys.
```
expede, line 36: 👍

spec/private-wnfs.md
```diff
@@ -33,7 +33,9 @@ The encrypted layer is intended to hide as much information as possible, while s
...
+
+Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multimap](https://en.wikipedia.org/wiki/Multimap). The HAMT MUST have a node-degree of 16, and MUST used saturated  saturated [namefilter](/spec/namefilter.md)s as the label. See [`rationale/hamt.md`](/rationale/hamt.md) for more information on parameter choice.
```
expede, line 38:
```suggestion
Ciphertext blocks MUST be stored as the leaves of the HAMT that encodes a [multimap](https://en.wikipedia.org/wiki/Multimap). The HAMT MUST have a node-degree of 16, and MUST used saturated [namefilter](/spec/namefilter.md)s as the label. See [`rationale/hamt.md`](/rationale/hamt.md) for more information on parameter choice.
```

rationale/encryption.md
```diff
@@ -0,0 +1,75 @@
+# File Encryption
+
+## Need
+
+A way to encrypt [`PrivateFile` and `PrivateDirectory` blocks](/spec/private-wnfs.md#31-cleartext-data) as well as externalized file content blocks.
+
+Even though authentication is not strictly necessary, because write access should be managed via namefilter certificates instead of knowledge-of-read-key, it is beneficial to have another layer of principals writing data with write access but without read access.
```
expede, line 7: I'm not 100% sure what you're trying to say in this paragraph. It's a run-on sentence, which may be part of the problem? Below is my attempt to break it up, but I'm still not totally certain if I captured what you were trying to say.

```suggestion
Note that authentication is not strictly necessary since write access is managed via PKI-signed certificates instead of knowledge of a secret (like a read key). However, it is often useful to have a layer of principals with write (but not read) access.
```

rationale/encryption.md
```diff
@@ -0,0 +1,75 @@
...
+## Choice
+
+We chose [AES-GCM] with per-message randomly generated 12-byte initialization vectors, because it is widely supported and implemented in hardware as of 2023.
+There are no known practical attacks on AES-GCM.
+
+## Alternatives
+
+### [AES-GCM-SIV]
+
+Compared to AES-GCM, this mode additionally protects against nonce-misuse via a synthetic initialization vector (SIV). To achieve this, AES-GCM-SIV sacrifices performance during encryption.
```
expede, line 18: In case it's helpful, I often point people at this post. It could make for a good footnote: https://soatok.blog/2020/07/12/comparison-of-symmetric-encryption-methods/

spec/private-wnfs.md
```diff
@@ -82,9 +84,9 @@ If the HAMT is used as the `PrivateForest` for WNFS, then the values stored SHOU
 
 ## 2.2 Ciphertext Files
 
-The encrypted file layer is a very thin enrichment of the data layer. In particular, it knows about namefilters as labels, and ciphertext blobs as being separate from the expanded namefilter inside the multi-valued entry.
+The encrypted file layer is a very thin enrichment of the data layer. In particular, it knows about namefilters as labels, and ciphertext blobs as being separate from the expanded namefilter inside the multivalued entry.
 
-![](./diagrams/hamt_leaves.svg)
+<img src="./diagrams/hamt_leaves.png" width="600">
```
expede, line 89: Is the schema for this in `2.1`? I don't see the namefilter in there, so may be missing something.

Mainly I'm curious if the namefilter is kept in the block as bytes, or if it's split out as its own node or something? I think maybe the ciphertexts being outside of the black boxes is not super clear to me in the current form 🤔

spec/private-wnfs.md
```diff
@@ -180,17 +207,17 @@ The block size MUST be at least 1 and at maximum $2^{18} - 28 = 262,116$ bytes,
...
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+| Initialization Vector |       |
++-+-+-+-+-+-+-+-+-+-+-+-+       |
+|                               :
+:         Encrypted Block       :
+:        (blockSize bytes)      |
+|                               |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|        Authentication Tag     |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```
expede, line 220: Very optional suggestion: consider using hex. I often see memory layout like this, and the decimal representation was less clear (and evidently was error prone)

```suggestion
 0 1 2 3 4 5 6 7 8 9 A B C D E F  (bytes)
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
| Initialization Vector |       |
+-+-+-+-+-+-+-+-+-+-+-+-+       |
|                               :
:         Encrypted Block       :
:        (blockSize bytes)      |
|                               |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
|        Authentication Tag     |
+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
```

spec/private-wnfs.md
```diff
@@ -212,13 +239,16 @@ See the section for [Read Hierarchy](#317-read-hierarchy) for more information a
 ### 3.1.6 Pointers & Keys
 
 Keys are always attached to pointers to some data.
+The pointer to that that data consists of the hashed namefilter, which is used as the label for a multivalue in the private forest and an accompanying CID to disambiguate which value in the multivalue is actually referred to.
 
-![](./diagrams/decryption_pointer.svg)
+<img src="./diagrams/decryption_pointer.png" width="400">
```
expede, line 244: It may be worth disambiguating that this a very abstract model, not actually how pointers relate to each other concretely

spec/private-wnfs.md
```diff
@@ -270,9 +300,9 @@ All algorithms MUST have access to a `PrivateForest` in their context.
 
 ## 4.1 Namefilter Hash Resolution
 
-`resolveHashedKey: Hash<Namefilter> -> (Namefilter, Array<Encrypted<PrivateNode>>)`
+`resolveHashedKey: Hash<Namefilter> -> (Namefilter, Array<Cid>)`
```
expede, line 303: Yeah I guess it's more straightforward to not try to mix levels 👍

expede (approved):
Looks great to me 👍 No blockers on the actual semantics, but a couple suggestions and clarifications that I hope are helpful! Great work 👏👏👏

spec/private-wnfs.md
expede: I assume it's a 128-bit authentication tag? If so, should that be reflected in the layout above?

matheus23: > the decimal representation was less clear (and evidently was error prone).

Yeah... agree. However, this type of diagram actually seems to be standardized in [RFC 2360](https://www.rfc-editor.org/rfc/rfc2360.html#section-3.1), which is why I went for it.

> I assume it's a 128-bit authentication tag? If so, should that be reflected in the layout above?

Isn't it? The "Authentication Tag" block is 16 bytes wide.

spec/private-wnfs.md
matheus23: > Is the schema for this in `2.1`?

Kind-of. The schema in 2.1 is generalized over any kind of labels and values. I keep going back-and-forth on this whether it's easier to be concrete or abstract in the beginning of 2.1.

> I don't see the namefilter in there, so may be missing something.

Technically the namefilter is instantiated as the `L` in the schema.

> Mainly I'm curious if the namefilter is kept in the block as bytes, or if it's split out as its own node or something?

The namefilter is kept in the block as bytes.
The ciphertexts are split out and referred to as CIDs.

---

I think abstracting the HAMT just doesn't work well in the spec. I'll make it WNFS-specific.

rationale/encryption.md
matheus23: Yep, I think it makes sense to link this!

rationale/encryption.md
matheus23: Thanks! I was missing some words: "another layer of *authentication for* principals".

I'll combine your suggestion and my fix :+1:

spec/notation.md
matheus23: Yeah, I'm not a big fan of `||` notation. I removed its usage. I was half using `concat(a, b)` and half using `a || b`. 
So yeah. Only `concat(a, b)` now.

spec/private-wnfs.md
matheus23: This is a good idea. Added a paragraph!

spec/private-wnfs.md
expede: LOL oops totally my bad. How does math work? 😅

PR #59, opened 2023-01-30
Non-Nested encryption implementation-guided specification fixes

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/non-nested-fixes/spec/private-wnfs.md)

* There's no need to store the "previous CID" for the header:
  You (should) need to know the previous revision's skip ratchet state to recover the previous revision's header's CID. Since the bare namefilter and inumber don't change between revision, you thus have all information to reconstruct the full header. (If you want to, you can encode it and thus derive the CID, as the header is encrypted deterministically using AES-KWP)
* Added a `contentCid` to share pointers.
  This prevents unnecessary guessing on the receiver side on which multivalue to decrypt. It also guarantees that whatever is seen at the time of sharing is seen at the time of receiving the share, even if there's concurrent writes in between (although these can be discovered by "searching latest" afterwards).
* Extracted out `PrivateRef` in the spec, like we did in the code


expede:
From our call, I'm leaving a comment: the diagram change makes sense! Also it would be good to put in prose the stuff about how this allows the user to very efficiently move inside a snapshot and only deal with the header for temporal lookups (move between snapshots). Because the arrow is now reversed, we may want to rethink the term "header", too!

matheus23:
> Also it would be good to put in prose the stuff about how this allows the user to very efficiently move inside a snapshot and only deal with the header for temporal lookups

I'm somewhat hesitant to put this into the spec. That is possible, but could be an attack vector if some node wrote inconsistent data into the `temporalKey` that's stored in a directory.
There exists an invariant, which is that the `TemporalKey` that's used to derive the `SnapshotKey` to encrypt the content node is the same as the key derived from the skip ratchet.
Due to there being an invariant, we need to think about what the source of truth should be in case of a conflict. That should be the skip ratchet. Thus when you read, you shouldn't assume that the temporal key you're reading out of a directory is actually the correct key, unless you've verified it against the skip ratchet it points to.

On the other hand - if you don't care about being lenient towards nodes with read & write access, then this optimization is possible?
I'm thinking I'll add a paragraph with that caveat.

expede:
I can't comment diretcly on image 🤦‍♀️ Totally happy to use TLDraw, but figured I'd see how quickly I could throw together an equivalent diagram in Mermaid (which is supported by GitHub flavoured markdown). You can do colours and shapes and stuff, but just as a quick sketch _for science:_

```mermaid
\flowchart LR
  Storage[Client Side Storage Secret Storage]
  TempKey[Temporal Key]
  SnapKey[Snapshot Key]

  subgraph privHeader [Private Node Header]
    skipRatchet[Skip Ratchet]
    bareNF[Bare Name Filter]
    inum[i-number]
  end

  subgraph privDir [Private Directory]
    metadata

    subgraph apps ["/Apps"]
      snapKeyApps[Snapshot Key]
      tempKeyApps[Temporal Key]
    end

    subgraph docs ["/Docs"]
      snapKeyDocs[Snapshot Key]
      tempKeyDocs[Temporal Key]
    end
  end

  Storage --> TempKey

  skipRatchet -- derives --> TempKey
  skipRatchet -- decrypts --> tempKeyApps
  skipRatchet -- decrypts --> tempKeyDocs


  TempKey -- decrypts --> privHeader
  TempKey -- SHA3 --> SnapKey

  SnapKey -- decrypts --> privDir
```

expede (approved):
Oh yeah this reads way better! 🤝 🎉

PR #62, opened 2023-02-22
Change enum variant signalling: Capsule Types

matheus23:
This PR changes the way enum (sum type) variants are signalled.
Before:
```typescript
type Node = File | Dir

type File = {
  type: "file"
  // ...
}

type Dir = {
  type: "dir"
  // ...
}
```

After:
```typescript
type Node = File | Dir

type File = {
  "file": { /* ... */ }
}

type Dir = {
  "dir": { /* ... */ }
}
```

I added some [:scroll: rationale](https://github.com/wnfs-wg/spec/blob/matheus23/enum-variant-signalling/rationale/encoding.md#enum-variant-signalling) about that in this PR, too.

We noticed this would make our life in rs-wnfs with serde *much* easier, and it's probably also the alternative that allows faster & more memory-efficient decoders in principle.

### DIFF
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -99,6 +99,8 @@ preimage like image
 IND-CCA
 multivalue like value
 PKI
+pre-defined
+erroring
 
 # rationale
 Combinatoric
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -103,6 +103,9 @@ PKI
 # rationale
 Combinatoric
 POS like ABC
+deserialize like serialize
+enum like structure
+javascript
 
 # other
 md
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -104,6 +104,7 @@ PKI
 Combinatoric
 POS like ABC
 deserialize like serialize
+deserialization like serialization
 enum like structure
 javascript
 
--- a/rationale/encoding.md
+++ b/rationale/encoding.md
@@ -1,14 +1,14 @@
 This documents our reasoning for current choices of encoding and potential alternatives.
 
-# Enum Variant Signalling
+# Enum Variant Signaling
 
 ## Need
 
-In various places, WNFS data structures are allowed to point to different types of data, e.g. a directory entry may point to yet another directory or a file instead. Encoding a datum that allows a deserialization procedure to find out which deserialization subprocedure to use is usually called signalling.
+In various places, WNFS data structures are allowed to point to different types of data, e.g. a directory entry may point to yet another directory or a file instead. Encoding a datum that allows a deserialization procedure to find out which deserialization sub-procedure to use is usually called signaling.
 
 ## Choice
 
-Signalling works by wrapping the data DAG-CBOR with an object containing a single key like this:
+Signaling works by wrapping the data DAG-CBOR with an object containing a single key like this:
 
 ```typescript
 type Node = File | Dir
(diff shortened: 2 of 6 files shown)


matheus23:
@expede had some good feedback on this, which I would try to summarize as this:

We need to be careful about data validation. With the `type: ...` approach, more libraries and developers are aware & thinking of the problem of malformed data in the form of a duplicate key.
With the capsule approach, the data can be malformed in more ways: There may be multiple keys in the capsule. Is it valid to have unrelated keys in the capsule? Is it valid to have multiple capsules in the same object?
For capsules, I think the correct answer is: They should be restricted. Only a single key in the capsule is allowed and it must be a known key or otherwise that's an error. This doesn't prevent extensibility: The object it refers to still has room for extension fields.

Parsing libraries will have varying support for this.
- A naive javascript or python implementation may just check a set of keys and see whether they're defined in the capsule, and if so - treat the object as of that variant. This allows all the foot guns mentioned above and may mean behaviour dependent on the order of keys checked. (But we don't want to impose an ordering of keys to check in the spec).
- Luckily, it turns out in rs-wnfs, where we're using the [serde](https://serde.rs/), the capsule parsing is restrictive just in the way I describe above: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=46f8b2cdbec9d76f263648ec7df43643

Anyway. I'll try to find a place in the spec adding a paragraph about strictness requirements for capsule parsing.

matheus23:
Actually, I'm realizing now that the same issues exist with `type: ...` variant signaling, the inconsistency is just hidden somewhere else:

```js
// What should this value be? "dir" or "file"?
JSON.parse(`{"type": "file", "type": "dir"}`).type
// The value may depend on the javascript runtime.
// A similar `parse` function may be used in python, but work differently!
```

---

I think the conclusion is: We need to specify that you should be careful about how you parse your data, otherwise that may cause inconsistencies between implementations.

expede:
@matheus23 

> Actually, I'm realizing now that the same issues exist with type: ... variant signaling, the inconsistency is just hidden somewhere else:

This is my other gripe with IPLD Schema assumtions, but technically an IPLD parser will always give you a consistent failure in these cases, becuase it's illegal syntactically. The capsule types are sometimes illegal semantically, which is way more ambiguous.

Not saying that you should pick one or the other, but the above is a factor.

matheus23:
Yeah that's right.
To be honest, I'm mostly not considering IPLD schemas/IPLD tooling too much yet, and instead thinking about more general purpose tooling, e.g. CBOR & JSON instead of DAG-CBOR and DAG-JSON, but perhaps I should. The reason that's important is (and I know you know this, but) duplicate keys are allowed in CBOR & JSON, but not in the dag- variants.

It's actually a really interesting insight, the "syntactically vs. semantically valid". I remember people musing about "adding sum type support" to JSON, and I guess that's what would help: It would make certain things invalid *syntactically*, so that capsules are easier to not mess up. Very interesting.

Yeah anyhow. My realization doesn't help here.

---

I realize that this *can* be a subtle foot gun, but it's an implementation-guided choice & helps us out quite a bit.
Looking at the [IPFS Schema docs](https://ipld.io/docs/schemas/features/representation-strategies/#union-keyed-representation), is it true that IPLD tooling will just 'do the right thing' here, as in, if you have a schema, it'll do the same thing with capsule types that serde is doing - disallowing multiple fields in the capsule?

expede:
Yeah 1000% agreed. Also agree that we should be very clear about the expected layout invariants in the docs regardless of IPLD Schema. We could use e.g. [canonicalized JSON](https://www.rfc-editor.org/rfc/rfc8785), but even then. It gets really hard with some of these more lax tools, so we probably need to write our own checking passes on these. It kind of hurts the argument that one layout will be much faster than the other 🤷‍♀️ Anyways, interesting to think about

matheus23:
I wrote a note about [:scroll: Validation](https://github.com/wnfs-wg/spec/blob/matheus23/enum-variant-signalling/spec/validation.md#notes-on-validation) in the spec and linked to it from a bunch of places where we're using capsule types.

Let me know what you think.

expede (approved):

PR #64, opened 2023-03-16
Exapand rationale against AES-GCM-SIV

expede:
[Preview 👀](https://github.com/wnfs-wg/spec/blob/expand-no-siv/rationale/encryption.md)

### DIFF
--- a/.github/workflows/words-to-ignore.txt
+++ b/.github/workflows/words-to-ignore.txt
@@ -1,106 +1,108 @@
 # Names
+Bluesky
+Holmgren like Alice
+Krüger like Alice
+Nyberg like Alice
+Philipp like Alice
 Zelenka like Alice
 expede
 matheus23
-Philipp like Alice
-Krüger like Alice
-Nyberg like Alice
-Holmgren like Alice
-Bluesky
 
 
 # Spec
-cleartext
-plaintext
-Cleartext
-sharding
-multimap
-untrusted
-namespace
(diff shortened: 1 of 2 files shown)


.github/workflows/words-to-ignore.txt
```diff
@@ -1,106 +1,108 @@
 # Names
```
expede, line 1: I had to add deduplication to this list (weird!) I also sorted it while I was in here to make stuff easier to find.

matheus23: Thank you! :)

matheus23 (approved):
Great summary :) I'm glad that we have the rationale files :grin:

.github/workflows/words-to-ignore.txt
matheus23: Yeah. If I write "deduplication" in this box, my browser also complains.
The spellchecker is using some standard american-english dictionary for hunspell. I wish there was a "programmer jargon"-specific dictionary though.

expede: It's fine. This will all get replaced by GPT eventually 😅

matheus23: *squinting*

That's actually not a bad idea. Makes me want to pull out some open-source LLM and attach it to my spellchecker. That's a self-nerd-snipe for another day though :grin:

PR #65, opened 2023-03-17
Name accumulators

matheus23:
[:scroll: Preview](https://github.com/wnfs-wg/spec/blob/matheus23/name-accumulators/spec/private-wnfs.md#4-private-path-representation)

This updates the spec to use RSA accumulators for write verification without read access instead of bloom filters (Namefilters).

- [x] PR Description
- [x] Deciding on exactly what to keep & what needs to be added
- [x] Rationale
- [x] Wording!


expede:
> [cla-bot] welcome to our community

Welcome to out commuuity, @matheus23 😉

matheus23:
@expede this was my desperate and failed attempt to get cla-bot to accept @actions-user

spec/diagrams/read_hierarchy.dot
```diff
@@ -46,7 +46,7 @@ digraph G {
             fillcolor = 5
         ]
 
-        external_namefilter [label = "Namefilter"]
+        external_nameaccum [label = "NameAccumulator"]
```
expede, line 49: I should maybe change this diagram to Mermaid — it's certainly not powerful enough to do lots of other kinds of diagrams, but ones like this one should work well

spec/private-wnfs.md
```diff
@@ -298,18 +308,85 @@ The skip ratchet is the single source of truth for generating the decryption key
...
+# 4 Private Path Representation
+
+WNFS is designed to support being stored with a service provider that can gate write access without the having read access to the file system and learning as little about the private file system hierarchy as possible.
+This presents a challenge: If the service provider can't read the submitted writes from clients, how should it decide which writes are valid and which are not?
+
+In WNFS, the solution is to represent private file system paths in a way that makes it possible to use them as the key in the private forest while leaking as little as possible about the path itself.
+This means the representation
+- must not leak information about the nodes that are part of the path,
+- must not be correlatable to other paths and
+- must not leak the amount of path segments inside the path.
```
expede, line 322: :eyes: Looks above. I guess no RFC2119 for now? I guess the style for a bunch of the doc would change, and this version does successfully communicates the ideas. Something to think about down the road at least.

spec/nameaccumulator.md
```diff
@@ -0,0 +1,130 @@
...
+
+This algorithm batches multiple batch elements proofs together. It combines the PoKCR (Proof of Knowledge of Co-prime Roots) from Section 3.3 of [this paper][IOP Batching Boneh] with the previously described PoKE* (see the last paragraph).
+
+The resulting proof is not fully succinct. It only manages to compress the number $Q$ from each `ElementsProof`, but still requires that you keep around all residues $r$ and $l$ hash counters as well as which accumulator states they are related to.
+
+The output consists of $Q^*$, the product of all numbers $Q$ from the `ElementsProof`s as well as all residues $r$ and all $l$ hash counters, mapped to the accumulator states they prove.
+
+
+[RSA acc og paper]: https://link.springer.com/content/pdf/10.1007/3-540-48285-7_24.pdf
+[IOP Batching Boneh]: https://eprint.iacr.org/2018/1188.pdf
```
expede, line 129: Yeah, Nessesary evil (or at least relative to our best understanding)

expede: Defining the exact formatting for how this separation happens is... outside of the spec? Or am I just not seeing it?

expede (approved):
It reads well to me! I've also recently put a bunch of this stuff back into my brain to discuss the "multiple batch" issue, so hearing from someone else if it's understandable may not be the worst thing in the world, but it gets a 👍 from me!

spec/private-wnfs.md
matheus23: Well, I kinda try to do RFC2119 everywhere, but in that case it's not like it's "on the implementer" to make the name accumulator *not* leak information. This is more like some prose on what the name accumulators are supposed to achieve.

spec/nameaccumulator.md
matheus23: There's no specified serialization format for the `MultiBatchProof` yet, no. I'm still thinking about what the best encoding may be.
It's actually very little data per key: You can get away with a 32-byte hash-of-namefilter + 16 bytes $r$ and ~1-2byte $l$ hash counter/nonce. In total that's ~40 bytes per key. You also additionally need a name accumulator (256 bytes) for each "allowed write name", but in most cases that's only 1 and usually at most 4?

So encoding these into a HAMT would be wasteful (at least with a normal "bucket size" of 5), since each leaf node would hold at max 16 * 5 = 80 entries = 3200 bytes. We'd like that to be more around 256kB.

Although perhaps I'll just abstract out the bucket size in our implementation and re-use our HAMT with a bucket size of ~256?

PR #66, opened 2023-07-12
Update rsa modulus endianess to match most protocols

matheus23:
Most protocols encode the RSA modulus in big-endian format.
E.g.
- [JWK spec](https://datatracker.ietf.org/doc/html/rfc7518#section-6.3.1.1) (see also [Section 2 "Terminology"](https://datatracker.ietf.org/doc/html/rfc7518#section-2) and `Base64urlUInt` encoding)
- ASN.1 DER RSAPublicKeys are encoded as ASN.1 DER `INTEGER`s. The best reference I found was [this stackoverflow answer](https://stackoverflow.com/a/52248395) :/ but I can confirm from working with the encoding in the past for ts-ucan.

1. This PR
2. #68 
3. #69
4. #70 
5. #71


spec/private-wnfs.md
```diff
@@ -465,7 +465,7 @@ function* shardLabels(key: Key, count: Uint64, name: NameAccumulator): Iterable<
 
 - `concat` denotes byte array concatenation,
 - `name` is the `NameAccumulator` from the private file's header,
-- `encode` is a function that maps a block index to a low-endian byte array encoding of a 64-bit unsigned integer.
+- `encode` is a function that maps a block index to a little-endian byte array encoding of a 64-bit unsigned integer.
```
expede, line 468: > Update rsa modulus endianess to match most protocols
> [...]
> Most protocols encode the RSA modulus in big-endian format.

This line says `little-endian`... so I'm confused?

expede: Looking at the next file, is the little-endian stuff part of the data format, or is it a implementation detail (libraries etc)?

matheus23: This is just me fixing a typo I saw on the way (low-endian vs. little-endian).
Some things in WNFS continue to be little-endian encoded for efficiency.
After this PR RSA nums are encoded as big-endian for consistency with other things (JWK/PKCS/etc.).

spec/shared-private-data.md
```diff
@@ -41,7 +41,7 @@ See the [example exchange partition layout](#exchange-partition-layout) for exam
 
 Individual device exchange keys are versioned. The protocol version is derived from the file name. Protocol version 1 exchange keys are named `v1.exchangekey`.
 
-The file's content is 256 bytes of the exchange key's RSA public modulus encoded as low-endian unsigned integer.
+The file's content is 256 bytes of the exchange key's RSA public modulus encoded as big-endian unsigned integer.
```
matheus23, line 44: This is the actual change :upside_down_face:

spec/private-wnfs.md
expede: Okay, so the actual data format that gets persisted to WNFS is always big endian, but sometimes we manually flip it for... library reasons? As you note elsewhere, a lot of _processors_ support both, and I would assume that anyoverhead there would be very fast, so the only reason that I can think of is that a library is doing direct bit fiddling?

matheus23: > the actual data format that gets persisted to WNFS is always big endian

No. Only RSA numbers are persisted as big endian. Everything else is little endian.

expede: Ah, okay. I had to think about the tradeoffs there a bit. I do remember us chatting about this a while back, and it does make sense even if it's nonobvious.

I wonder if the text should make this fact more salient? It's a big difference and easy to miss. Maybe a sentence that this is different from the test of the algos?

expede: ...in the section that you highlighted below, of course (not here)

expede (approved):
Just the one suggestion, but LGTM! 🚢

spec/private-wnfs.md
matheus23: I'm coming back to this after being confused in implementations, I think you're right @expede, it should just all be the same endianess. So, just to summarize:
- Having both big & little endianess in the same protocol sucks. rs-wnfs had the endianess of some things wrong now for a while without me noticing. That's *bad*.
- Most protocols that deal with RSA are big-endian. And because e.g. we're importing RSA moduli from the outside sometimes, we should *also* be big-endian. More concretely: I don't want an implementation to import a WebCrypto-generated RSA modulus (which is big-endian) and rs-wnfs interpreting that as little-endian and suddenly it's not a safe RSA modulus at all!
- The performance difference between little & big endianess was in fact so minuscule, I didn't notice I used the wrong endianess for a while.

I'll create a PR to update everything to be big-endian.

PR #69, opened 2023-07-17
Specify Blake3's `derive_key` algorithm instead of SHA3

matheus23:
- Use Blake3 instead of SHA3 everywhere
- Use the `derive_key` algorithm from blake3 (see [the specification](https://github.com/BLAKE3-team/BLAKE3-specs/blob/master/blake3.pdf)) if applicable
- Introduce some domain separation strings to use with `derive_key`
- Adjust the way external file content blocks are linked to skip a lot of work.

:exclamation: **merge order** :exclamation: 
1. #66 
2. #68 
3. This PR
4. #70 
5. #71

PR #68, opened 2023-07-17
Remove `namefilter.md` and add `NameAccumulator` verification algorithm

matheus23:
And a couple of small fixes, too.

:exclamation: **merge order** :exclamation: 
1. #66 
2. This PR
3. #69
4. #70 
5. #71

PR #70, opened 2023-07-18
Switch from `AES-GCM` to `XChaCha20-Poly1305`

matheus23:
- Updated the notation to use `Encrypted<>` instead of `AesGcm<>` and `KeyWrapped<>` instead of `AesKw<>`. This is more in line with how `Hash<>` was named.
- Updated the text to refer to a 24 byte IV, and updated the byte threshold per-block to 262,104 bytes.
- Updated the ascii diagram

1. #66 
2. #68 
3. #69
4. This PR
5. #71


rationale/encryption.md
```diff
@@ -11,12 +11,25 @@ Note that authentication is not strictly necessary since write access is managed
...
 
 ## Alternatives
 
+### [ChaCha20-Poly1305]
+
+This is the cipher that [XChaCha20-Poly1305] is based on, but has a smaller nonce size. Although it is more popular, we believe the extended nonce is worth prioritizing over popularity.
+
+These ciphers run at roughly 4 times the throughput in software compared to AES-GCM and are comparatively easier to implement as constant-time algorithms.
+However, AES-GCM has wider hardware support at the moment.
+In the future, it's likely we'll reconsider these ciphers.
```
expede, line 26: I'd maybe put a line in here about the reason (e.g. post-quantum crypto). Without context, right now I'd read this as either us having unstable APIs and/or low confidence in these algorithms, which is not the case.

spec/notation.md
```diff
@@ -19,16 +19,16 @@ The CID's multicodec is meant to be inferred automatically from its type paramet
 
 The *actual* representation of `Cid` is meant to be a byte representation for CIDs, and depends on the context. E.g. if `Cid` is found somewhere within a `Cbor<>` block, then it uses the DAG-CBOR-native way of encoding CIDs as CBOR objects with a `42` tag and a binary representation of a CID.
 
-## `AesGcm<Data>`
+## `Encrypted<Data>`
 
-This represents a ciphertext that's marked with the data type `Data` that it needs to decrypt to. For example `AesGcm<Ratchet>` represents a byte string that would decrypt to a value of type `Ratchet` using [AES-GCM](https://csrc.nist.gov/publications/detail/sp/800-38d/final) with 256-bit keys.
-The ciphertexts produced by `AesGcm<>` have a 12 byte initialization vector prepended and the 16 byte authentication tag appended.
+This represents a ciphertext that's marked with the data type `Data` that it needs to decrypt to. For example `Encrypted<Ratchet>` represents a byte string that would decrypt to a value of type `Ratchet` using [XChaCha20-Poly1305] with 256-bit keys.
+The ciphertexts produced by `Encrypted<>` have a 24 byte initialization vector prepended and the 16 byte authentication tag appended.
```
expede, line 25: Interesting, is `Encrypted<>` a Rust convention? Obvs as a Haskeller, I expect to see a type variable in there (`Encrypted<T>`)

expede: Actually maybe it's because I see the brakcets? You can write `Maybe` (of kind `* -> *`), so perhaps it's not that weird.

spec/private-wnfs.md
```diff
@@ -25,23 +25,23 @@ These all form graphs, where the nodes and links have different meanings per lay
 
 # 2 Encrypted Layer
 
-The encrypted layer hides the structure of the file system that it contains. The data MUST be placed into a flat namespace — in this case a [Merklized](https://en.wikipedia.org/wiki/Merkle_tree) [hash array mapped tire (HAMT)](https://en.wikipedia.org/wiki/Hash_array_mapped_trie). The root node of the resulting HAMT plays a very different role from a file system root: it "merely" anchors this flat namespace, and is otherwise unrelated to the file system. The file system structure will be ["rediscovered" in the decrypted layer (§3)](#3-decrypted-layer).
+The encrypted layer hides the structure of the file system that it contains. The data MUST be placed into a flat namespace — in this case a [Merklized][Merkle Tree] hash array mapped trie ([HAMT]). The root node of the resulting HAMT plays a very different role from a file system root: it "merely" anchors this flat namespace, and is otherwise unrelated to the file system. The file system structure will be ["rediscovered" in the decrypted layer (§3)](#3-decrypted-layer).
```
expede, line 28: 🤔 I could be wrong, but aren't we using the Filecoin IPLD HAMT which is modified to use some CHAMP features?

expede (approved):
A couple minor questions and suggestions, but otherwise looks good :)

spec/private-wnfs.md
matheus23: > aren't we using the Filecoin IPLD HAMT
Yeah we are.

> which is modified to use some CHAMP features?
Well, yeah. The champ feature that it uses is that it canonicalizes nodes on remove. E.g. if you remove enough keys from an internal node, the node gets merged with its parent.

However, the encoding is pretty much a HAMT. I.e. there's only a single bitmap in internal nodes, not two (as would be the case for CHAMPs).

rationale/encryption.md
matheus23: Uuuuh. That's a copy-paste error from when AES-GCM was still the WNFS algorithm. I removed it :+1:

Issue #50, opened 2022-11-14
Link to external file content from metadata


matheus23:
Goal: Solve issues raised in discussions #18 and #49.
The use case is a file system that supports more than just a single byte array as a file's data. E.g. storing thumbnails of images.

On the public file system this can be supported with adding `CID`s that point to more (unixfs) data into metadata, but on the private file system this doesn't work as expected: The CID isn't "discoverable", it needs to be added to the private forest to be syncable. It also needs to be encrypted.

Thus the need for linking to external encrypted bytes from the metadata.

This would be pretty much supporting the same type that `.content` supports, but within metadata.
For the private side that means that metadata may include content that looks like this:

```ts
type ExternalContent = {
  key: Key
  blockSize: Uint64 // in bytes, at max 262132
  blockCount: Uint64
}
```

For example, this could be a file's metadata represented as JSON with a link to some thumbnail:
```json
{
  "createdAt": 9999999,
  "modifiedAt": 9999999,
  "thumbnail": {
      "key": "<bytes>",
      "blockSize": 262132,
      "blockCount": 3
  }
}
```

We should also specify that implementations support operations like `read_metadata_linked(file, metadataKey: string): Bytes`.

matheus23:
One remaining question to resolve: What does should happen if some process writes to the file that doesn't know about the mechanics behind that metadata.

An example:
1. I add a picture on my capyloon device. It creates a file with a thumbnail stored in the metadata
2. I open up that picture on e.g. some kind of web-drive viewer and change the image. The web-drive editor doesn't know about thumbnail metadata.

What should happen? Should the new revision inherit all metadata from the previous revision or should the new revision not have metadata associated?

---

Perhaps it makes sense to have two types of metadata:
1. Metadata on *the specific version of a file*. Every revision gets a reset of this metadata. Let's call this "revision metadata"
2. Metadata on *the file name/identity*. New revisions inherit metadata from old revisions. Let's just call this "file metadata".

Updating a file clears revision metadata, but keeps file metadata.
Updating file or revision metadata will create a new revision with updated metadata. If only the metadata changes, even the revision metadata gets inherited!

These things would/could be file metadata:
- Tags
- mime type(?)
- creation time
These would/could be revision metadata:
- modification time
- associated thumbnail
- person/process that created this revision

jessmartin:
[...]

PR #71, opened 2023-07-18
Add `baseName` to `ExternalContent`

matheus23:
We explicitly add `baseName` to the `ExternalContent` record.
Usually `baseName = file.header.name.add(externalContent.key)`, but in some cases it's nice to for it to be something different.
This is nice, because it makes `ExternalContent` self-contained. You only need the information from `ExternalContent` to be able to read the whole content bytes.
This allows us to potentially use it in other places (e.g. thinking of #50), as well as copying it over from other `PrivateFile`s, in case a file gets moved or copied, without having to re-encrypt a bunch of data.

1. #66 
2. #68 
3. #69
4. #70 
5. This PR


spec/private-wnfs.md
```diff
@@ -211,7 +212,9 @@ Private file content has two variants: inlined or externalized. Externalized con
 
 #### 3.1.4.1 Externalized Content
```
expede, line 213: Nit: it would be good to expand section `3.1.4`. I've had to re-load a bunch of context about the difference between inline and external content, which means that it will very likely be confusing to someone new coming in.

expede: Specifically some examples of when you'd want one or the other would be helpful

spec/private-wnfs.md
```diff
@@ -211,7 +212,9 @@ Private file content has two variants: inlined or externalized. Externalized con
...
 
-Since external content blocks are separate from the header, they MUST have a unique `NameAccumulator` derived from a random key (to avoid forcing lookups to go through the header). If the key were derived from the header's key, then the file would be re-encrypted e.g. every time the metadata changed. See [sharded file content access algorithm](#44-sharded-file-content-access) for more detail.
+Since external content blocks are separate from the header, they each MUST have a `NameAccumulator` that is different than the file's `NameAccumulator`. We allow these names to have an arbitrary base. For the normal case, the base name is RECOMMENDED to be the file's name with the externalized content's encryption `key`, hashed to a prime, added to it as a name segment.
```
expede, line 215: Is "base" here a term defined elsewhere? It may be worth linking to it, or using a more descriptive term here

expede: Perhaps it's this field?

```
baseName: NameAccumulator
```

expede: > the file's name with

The nameaccumulator, or the human readable name? The `with` here is concatenation?

spec/private-wnfs.md
matheus23: I tried to expand the section a little bit :eyes:
LMK what you think

spec/private-wnfs.md
matheus23: The `NameAccumulator`. I tried to make this more explicit by saying "file's `name` from its header".
LMK if that's better :)

spec/private-wnfs.md
expede: Super great! Thanks :)

expede (approved):
Looks good! 👏

